Issue #6, opened 2026-08-19
Snapshot RateMyProfessors ratings nightly


EnesYilmazcode:
RateMyProfessors sends no CORS headers, so the browser cannot call it and a naive design would need a server. It does not need one: the full Ohio State roster is bulk readable (8366 professors, school id `U2Nob29sLTcyNA==`), and ratings barely move day to day.

So snapshot the roster on a schedule and commit the result as a static file. GitHub Pages then serves it like any other asset, and the site stays serverless.

**Scope**
- `scripts/fetch-ratings.mjs` pages the full OSU roster and writes `data/ratings.json`
- Store name, department, average rating, rating count, difficulty, would-take-again, and RMP id
- `.github/workflows/ratings.yml` running nightly and on manual dispatch, committing only when the file actually changes
- Fail loudly if the roster comes back suspiciously small, so a silent upstream change cannot quietly empty the file

**Acceptance criteria**
- Workflow runs green and produces a committed `data/ratings.json`
- Rerunning with no upstream change produces no commit

Issue #2, opened 2026-08-19
Document the OSU class API


EnesYilmazcode:
The old site (osucoursesearch.org) is down and its author is unreachable. It solved this problem with a Selenium grid scraping PeopleSoft plus a daily text dump from a personal faculty page. Neither is necessary: OSU runs a public JSON API that powers classes.osu.edu.

Write it down in `docs/osu-api.md` so this project never has to rediscover it.

**What to document**
- Base: `https://content.osu.edu/v2`
- `GET /classes/searchableTermsV2` returns the searchable terms
- `GET /classes/search?q=&campus=col&term=&p=` returns courses with nested sections
- Response shape, including that `meetings[].instructors[]` carries `displayName`, `role`, and `email`
- No auth, and `Access-Control-Allow-Origin: *`, so a browser can call it directly
- Known limits: broad queries cap at 10000 results (50 pages of 200), so full pulls must iterate by subject
- `GET /classes/availability` returns 503 on every parameter shape tried, so treat it as retired
- The search response has `enrollmentTotal` but no enrollment cap, so do not render a percent-full meter

**Acceptance criteria**
- Someone reading only this doc could rebuild the client from scratch

Issue #4, opened 2026-08-19
Build search and course results


EnesYilmazcode:
One search field, because the API takes one `q` and already matches subject, catalog number, title, and instructor name. Verified: `q=CSE 2221` finds the course and `q=Bucci` finds that instructor's 17 classes.

**Scope**
- Single search input plus term selector
- Render each result as a course header (subject, catalog number, title, credits) with its sections beneath
- Section row shows class number, component, meeting days and times, building and room, instruction mode
- Empty state, loading state, and no-results state that each say something useful

**Acceptance criteria**
- Searching `CSE 2221` for Autumn 2026 returns the course with all its sections
- Searching a professor's last name returns their classes

PR #10, opened 2026-08-19
Set up repo scaffolding

### LINKED ISSUE #1: Set up repo scaffolding
Lay down the project structure before any feature work so every later PR has an obvious home.

**Scope**
- `.gitignore` for Node and local scratch files
- MIT `LICENSE`
- `README.md` explaining what Finder does, where the data comes from, and how to run it locally
- Directory layout: `index.html` at root (GitHub Pages serves root), `scripts/` for the ratings job, `docs/` for research notes

**Acceptance criteria**
- Repo has a real README that a stranger can follow
- `git clone` plus opening `index.html` does something sensible, even if it is just an empty shell

EnesYilmazcode:
Closes #1

Lays down the project structure so every later PR has an obvious home.

## What's here
- `.gitignore` covering Node, editor dirs, and local scratch
- MIT `LICENSE`
- `README.md` explaining what Finder does, why it exists, the serverless architecture, and how to run it locally
- Placeholder `index.html` so the repo loads in a browser before the real UI lands

## Notes for review
- The README links to `docs/osu-api.md`, which does not exist until #2 merges. Deliberate, since #2 is the next PR, but worth flagging.
- `scripts/` and `data/` are not in this PR. Git does not track empty directories and inventing placeholder files to force them in would be noise. They arrive with #6.

## Verification
Served the folder with `python -m http.server` and confirmed the page loads and renders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+node_modules/
+.DS_Store
+Thumbs.db
+*.log
+.env
+.env.*
+.vscode/
+.idea/
+scratch/
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Enes Yilmaz
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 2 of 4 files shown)


EnesYilmazcode (commented):
Reviewed the diff. Scaffolding is right, and the architecture note in the README is the part worth keeping, since it records *why* there is no server rather than just stating that there isn't one.

Two instances of the same defect, both documentation promising things that do not exist yet:

1. `README.md:37` links `docs/osu-api.md`. Missing until #2.
2. `README.md:54` tells the reader to run `node scripts/fetch-ratings.mjs`. Missing until #6.

The PR description flags the first but not the second, and the second is worse: a broken link is obvious to a reader, whereas a documented command that just fails is a person following instructions and hitting an error. Verified both are absent on this branch.

Not blocking. #2 is the next PR and lands the doc, #6 lands the script, and both sit in the same milestone. Merging on the condition they close in that order. If either slips, the README should lose the reference rather than keep pointing at nothing.

Nit, non-blocking: `index.html` ships with no styling, so the placeholder renders as raw Times New Roman. It is at least honest about being a placeholder, so leaving it for #4.

PR #11, opened 2026-08-19
Document the OSU class API

### LINKED ISSUE #2: Document the OSU class API

EnesYilmazcode:
Closes #2

Records the API that replaces the old tool's Selenium plus text-dump pipeline, so this never has to be rediscovered.

## What's here
`docs/osu-api.md` covering:
- How the endpoints were found (they are in `osu-mobile.js`, not the page HTML)
- `searchableTermsV2` and `search`, with parameters and a real response
- The nesting quirk that instructors hang off `meetings[]`, not off the section
- Term code format (`1268` = Autumn 2026)
- Measured limits: the 10,000 result ceiling, the absent enrollment cap, late future terms
- `availability` is retired, 503 on every shape tried
- Why PeopleSoft scraping and Barrett's schedule were rejected, with the head-to-head CSE numbers

## Verification
Every endpoint, header, and limit in the doc was called directly rather than assumed.

## Known problem with this PR
The comparison table's last row is wrong and I am flagging it in review below rather than quietly amending, since the underlying measurement turned out to be unstable and that instability is itself worth documenting.

### DIFF
--- a/docs/osu-api.md
+++ b/docs/osu-api.md
@@ -1,0 +1,179 @@
+# The OSU class API
+
+Ohio State runs a public JSON API that serves course, section, instructor, and
+enrollment data. It is undocumented publicly, requires no authentication, and is
+what powers [classes.osu.edu](https://classes.osu.edu). Everything Finder shows
+comes from it.
+
+This document exists so the project never has to rediscover it.
+
+## How it was found
+
+`classes.osu.edu` is an AngularJS shell. The page HTML is 3.5 KB and contains no
+API references. The endpoints are in the bundle it loads, `osu-mobile.js`, which
+names a base of `https://content.osu.edu/v2` and three paths under `/classes`.
+
+## Base
+
+```
+https://content.osu.edu/v2
+```
+
+No API key, no login, no session cookie. Responses include:
+
+```
+Access-Control-Allow-Origin: *
+Access-Control-Allow-Methods: GET, OPTIONS, HEAD
+```
+
+That header is the reason Finder needs no backend. A browser can call this API
+directly from any origin, including a GitHub Pages site.
+
+## Endpoints
+
+### GET /classes/searchableTermsV2
+
+Returns the terms currently open to search. Use it to populate a term selector
+rather than hardcoding, because the set changes as terms roll over.
+
+```json
+{"data":{"data":[
+  {"strm":"1268","descr":"Autumn 2026","classSearch":"Y",
+   "startDate":"2026-02-09","endDate":"2027-01-31"}
+]}}
+```
+
+Measured on 2026-08-18, this returned exactly three terms: Spring 2026 (`1262`),
+Summer 2026 (`1264`), and Autumn 2026 (`1268`).
+
+### GET /classes/search
+
+The main endpoint.
+
+| Parameter | Notes |
+|---|---|
+| `q` | Free text. Matches subject, catalog number, course title, and instructor name |
+| `campus` | `col` for Columbus |
+| `term` | A `strm` code, for example `1268` |
+| `p` | 1-based page number |
+
+`q` matching instructor names is worth calling out, because it means searching by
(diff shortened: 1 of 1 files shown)


EnesYilmazcode (commented):
**Review: changes needed before merge.** (Filed as a comment because GitHub will not let an author formally request changes on their own PR.)

The doc is the right artifact and most of it holds up, but re-running the measurements before approving turned up two problems, one of which invalidates a table row and one of which is a genuine API limit the doc misses entirely.

## 1. The comparison table's last row conflates two different measurements

The row reads:

> Sections where it names an instructor and the other does not: API **649**, Barrett **0**

Barrett's `0` is correct and is the row that matters. The API's `649` is not the same quantity. 649 was the count of sections the API has that Barrett lacks *at all*, which is a coverage difference, not an instructor-naming difference. Putting them side by side in one row implies they were measured the same way. They were not.

Measured properly, restricted to the 388 sections both sources actually share:

```
IN OVERLAP: Barrett names an instructor, API blank  ->  0
IN OVERLAP: API names an instructor, Barrett blank  ->  0
```

So on shared sections the two agree completely, and the API's real advantage is coverage, not naming. The honest claim is narrower than the one written, and it is still more than enough to justify the decision.

## 2. Paged search is not deterministic, and the doc does not say so

This is the more serious finding. Section counts moved between runs, so I pulled the identical query three times back to back:

```
run 1: 1001 CSE sections    totalPages=7
run 2: 1010 CSE sections    totalPages=7
run 3: 1001 CSE sections    totalPages=7

intersection of all 3:  947
union of all 3:        1064
missing from run 1:      63
```

Same query, same term, same page count, and roughly 6% of sections differ per pull. Any single multi-page pull silently misses sections. Every hardcoded figure in the comparison table inherits that error bar, which is why `1046` in the table, `992` from an earlier run, and `965` from another are all equally "correct".

This has a design consequence beyond the doc: **Finder must not depend on exhaustive multi-page pulls.** A user searching `CSE 2221` fits on one page and is unaffected. A future feature that enumerates a whole subject would be quietly lossy. That constraint belongs in the doc so it is not relearned later as a bug.

## Requested changes
- Rewrite the last table row to compare like with like, or split coverage and naming into separate rows
- State section counts as approximate, with the reason
- Add a limits entry for non-deterministic paging, including the three-run evidence and the "keep queries narrow" consequence

Everything else, including the freshness analysis and the rejection of PeopleSoft, stands as written.

EnesYilmazcode:
Review addressed in `1dc7a3e`.

- Split the conflated row into a coverage table and a separate naming table scoped to the 388 shared sections. The decisive claim is now stated narrowly and correctly: on shared sections, neither source names an instructor the other misses, and the API wins on coverage rather than naming.
- Section counts now read as approximate, with the reason attached.
- Added a limits entry for non-deterministic paging, with the three-run evidence and the "keep queries narrow enough to fit on one page" constraint, so #4 does not get designed around an assumption that does not hold.

Merging.

Issue #9, opened 2026-08-19
Responsive and accessibility pass


EnesYilmazcode:
Registration happens on a phone at 7am during an enrollment window, so mobile is the primary case, not an afterthought.

**Scope**
- Layout works down to 360px wide
- Visible keyboard focus on every interactive element
- Search field and results wired up for screen readers, with results announced on change
- Honor `prefers-reduced-motion`
- Check color contrast, especially the status colors

**Acceptance criteria**
- Usable end to end with keyboard only
- No horizontal scrolling at 360px

PR #12, opened 2026-08-19
Add API client and term selector

### LINKED ISSUE #3: Add API client and term selector
Build the thin layer everything else sits on.

**Scope**
- `fetchTerms()` calling `searchableTermsV2`
- `searchClasses({q, term, page})` calling `/classes/search`
- Term selector in the UI, defaulting to Autumn 2026 (`1268`)
- Handle network failure with a message that says what broke and what to do

**Acceptance criteria**
- Term selector populates from the live API, not a hardcoded list
- Autumn 2026 is selected on load

EnesYilmazcode:
Closes #3

The thin layer everything else sits on.

## What's here
- `js/api.js` with `fetchTerms()`, `searchClasses()`, and an `ApiError` type
- `js/app.js` populating the term selector on load
- Term selector wired into `index.html`

## Decisions worth reviewing

**The default term is derived, not hardcoded.** The issue said default to Autumn 2026 (`1268`). Hardcoding that would be wrong the moment the term ends, so `termCodeFor()` computes the code from today's date using OSU's `strm` format (`1` + two-digit year + `2`/`4`/`8` for Spring/Summer/Autumn) and `defaultTerm()` falls back to the newest searchable term if the computed one is not offered. Today that yields `1268`, satisfying the acceptance criterion without the date bomb.

**Errors are written for a student, not a developer.** A failed fetch says Ohio State's course service could not be reached and suggests retrying, rather than surfacing a `TypeError`. Unexpected errors still hit the console.

## Verification
Ran against the live API:

```
termCodeFor 2026-08-18 -> 1268
termCodeFor 2026-03-01 -> 1262
termCodeFor 2026-06-01 -> 1264
terms: 1268 Autumn 2026 | 1264 Summer 2026 | 1262 Spring 2026
default: 1268 Autumn 2026
search CSE 2221 -> totalItems 1281, courses 34, pages 7
```

## Flagging for #4
That last line matters. `q=CSE 2221` returns **1281 items across 7 pages**, so the API matches loosely rather than treating the query as one course. Page 1 alone has 34 courses. #4 will need client-side ranking so the course you actually typed comes first, and it should not assume a query maps to a single course.

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,8 +10,14 @@
   <main>
     <h1>Finder</h1>
     <p>Find who teaches your classes before everyone else.</p>
-    <p>The search interface is not built yet. Follow along in
-       <a href="https://github.com/EnesYilmazcode/Finder/milestone/1">milestone v1</a>.</p>
+
+    <label for="term">Term</label>
+    <select id="term" disabled></select>
+
+    <p id="status" role="status" hidden></p>
+
+    <p>Search lands in <a href="https://github.com/EnesYilmazcode/Finder/issues/4">#4</a>.</p>
   </main>
+  <script type="module" src="js/app.js"></script>
 </body>
 </html>
--- a/js/api.js
+++ b/js/api.js
@@ -2,6 +2,7 @@
 
 const BASE = "https://content.osu.edu/v2";
 const CAMPUS = "col";
+const TIMEOUT_MS = 12000;
 
 export class ApiError extends Error {
   constructor(message, { status = null, cause = null } = {}) {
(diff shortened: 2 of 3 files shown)


EnesYilmazcode (commented):
**Review: changes needed before merge.**

The derived term default is the right call and the reasoning holds. Two real bugs in `js/api.js` though, one of which will produce a 500 the first time someone submits an empty search.

## 1. Empty query is silently dropped, and the API 500s on it

`getJson()` filters out empty-string parameters:

```js
if (v !== undefined && v !== null && v !== "") url.searchParams.set(k, v);
```

`searchClasses()` passes `q: q ?? ""`, so an empty search sends **no `q` parameter at all** rather than an empty one. Those are not equivalent upstream:

```
GET /classes/search?campus=col&term=1268&p=1
  -> 500  "Cannot read properties of undefined (reading 'replace')"

GET /classes/search?q=&campus=col&term=1268&p=1
  -> 200  totalItems: 10000
```

The API dereferences `q` without checking it exists. So the omission turns a harmless empty search into a server error, and the user gets "Ohio State's course service returned an error (500)" for pressing enter on an empty box.

The empty-string filter is still right for optional parameters generally. `q` just is not optional. Either pass it unconditionally or drop the `!== ""` test for it.

## 2. No request timeout, so a hung fetch hangs the UI permanently

`fetch()` is called with no `AbortController`, and browsers will wait a long time before giving up. `init()` sets "Loading terms..." and only clears it in `try`/`catch`, so a request that neither resolves nor rejects leaves the selector disabled and the message on screen with no way forward.

This is not hypothetical for this project. Campus networks and captive portals are exactly where sockets hang open, and the target user is on a phone during a registration window. Add an `AbortController` with a sensible deadline and treat the abort as a normal `ApiError`.

## Also worth noting, not blocking
- `#status` uses `role="status"` but is toggled with the `hidden` attribute. Assistive tech often will not announce content inserted into a region that was hidden at the time. Keeping the element present and swapping only its text is more reliable. Leaving it for #9, but flagging so it is not forgotten.
- `status.dataset.kind` is set but nothing consumes it yet. Fine as long as #4 styles it.

Blocking on 1 and 2. Both are small.

EnesYilmazcode:
Both fixed.

**1. Empty query.** `getJson()` now sends empty-string params instead of dropping them, with a comment recording why, since the reason lives upstream and is not guessable from this file. Regression tested against the live API:

```
empty query  -> 10000 items, 87 courses   (was a 500)
normal query -> 1281 items                (unchanged)
```

**2. Timeout.** Added a 12 second `AbortController` deadline, with abort surfaced as its own `ApiError` message rather than being lumped in with connection failure, since "taking too long" and "cannot reach" want different reactions from the user. `clearTimeout` runs in `finally` so the timer never leaks on the success path.

The `role="status"` announcement issue is left for #9 as noted.

Merging.

Issue #13, opened 2026-08-19
Get real per-section seat counts


EnesYilmazcode:
The OSU search endpoint reports enrollment that is wrong at the section level, so Finder currently shows no seat data at all. That is the honest choice, but it removes the "can I actually get into this" signal, which is most of why someone checks a section in the first place.

## The problem

`enrollmentTotal` and `enrollmentStatus` come back identical for every section of a course. Checked across five courses:

```
CSE 2221         22 sections   distinct enrollmentTotal: [32]   statuses: ["Open"]
MATH 1151        74 sections   distinct enrollmentTotal: [97]   statuses: ["Open"]
ENGLISH 1110.01  42 sections   distinct enrollmentTotal: [21]   statuses: ["Closed"]
PSYCH 1100       23 sections   distinct enrollmentTotal: [60]   statuses: ["Closed"]
CSE 3241         11 sections   distinct enrollmentTotal: [35]   statuses: ["Open"]
```

One value, repeated onto every section. It is not per-section data.

Worse, it is wrong in the direction that hurts. Barrett's schedule has real per-section numbers for the same class numbers:

```
class   Barrett          API
5168    32/40            32 enrolled, Open      <- agrees
4827    40/40  FULL      32 enrolled, Open      <- API says open, section is full
4831    41/40  +1 wait   32 enrolled, Open      <- API says open, section is over cap
```

A student trusting Finder would sign up for a 9:10am section that has been full for weeks. Shipping that is worse than showing nothing, which is why #4 removed it.

## The fix

Barrett's schedule carries accurate enrolled counts **and** the enrollment limit, which the OSU API does not expose at all. But `asc.ohio-state.edu` sends no CORS headers, so a browser cannot fetch it.

Same shape as the ratings problem in #6, so the same solution: snapshot it in CI.

- Add a scheduled job that fetches Barrett's per-subject text files and parses them into `data/seats.json`, keyed by class number
- The format is fixed-width. Column map, verified against Autumn 2026:
  ```
  col  5    9      20-30   31    33-42      49-53      58-68  70-76  78-88    89-92  94+
       SUBJ NUMBER class#  type  (autoenr)  M T W R F  time   room   enr/lim  +wait  instructor
  ```
- Parse with a residue check: every character on a section line must land in a known field, or the line is flagged rather than silently dropped
- Merge onto sections client-side by class number, and show `32/40` style counts plus a full/waitlisted state

## Caveats to carry into the implementation
[...]

PR #14, opened 2026-08-19
Build search and course results

### LINKED ISSUE #4: Build search and course results

EnesYilmazcode:
Closes #4

Finder now does the thing it exists to do: type a course, see its sections and who teaches them.

## What's here
- `css/finder.css`, design tokens with a dark mode and a reduced-motion guard
- `js/format.js`, meeting days, times, places, credits, instructor dedupe
- `js/rank.js`, merges duplicate course records and reorders results
- `js/render.js`, builds DOM nodes, never HTML strings
- `js/app.js`, wiring, URL state so a search is shareable
- Real markup in `index.html`

## Ranking was necessary, not polish
`q=CSE 2221` returns **1281 items across 7 pages**, and page one carries CSE 2231, CSE 2321 and CSE 5022 before you find what you typed. The API also splits some courses across duplicate records, so CSE 5052 appeared twice with one section each.

After merge and rank:

```
q="CSE 2221"   raw= 31  merged= 30  top3: CSE 2221, CSE 2231, CSE 5022
q="MATH 1151"  raw= 16  merged= 16  top3: MATH 1151, MATH 1131, MATH 1172
q="Bucci"      raw=  9  merged=  9  top3: CSE 2231, CSE 5023, CSE 2221
q="2221"       raw= 36  merged= 34  top3: CSE 2221, MUSIC 2221, ANIMSCI 2221
```

## One thing removed on purpose

I built seat counts, then found the data is wrong, so I took them back out. `enrollmentTotal` is one number per course repeated onto every section:

```
class   Barrett          API
5168    32/40            32 enrolled, Open      agrees
4827    40/40  FULL      32 enrolled, Open      API says open, section is full
4831    41/40  +1 wait   32 enrolled, Open      API says open, section is over cap
```

Shipping that would send someone to a section that filled weeks ago. Full write-up and the fix are in #13.

## Verification
Ran against the live API and rendered in headless Chrome. Responsive measured at six widths:

```
 320px  scrollWidth=305  clientWidth=305  overflow=no  spills=none
 360px  scrollWidth=345  clientWidth=345  overflow=no  spills=none
 390px  scrollWidth=375  clientWidth=375  overflow=no  spills=none
 414px  scrollWidth=399  clientWidth=399  overflow=no  spills=none
 768px  scrollWidth=753  clientWidth=753  overflow=no  spills=none
1100px  scrollWidth=1085 clientWidth=1085 overflow=no  spills=none
```

Note for anyone reproducing this: Chrome's `--headless=new --window-size` is ignored for layout, `innerWidth` stays at 500 no matter what you pass, and the screenshot then gets cropped so narrow captures look broken when they are not. Measure in a sized iframe instead.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -1,0 +1,293 @@
+/* Finder. Scarlet is structural here, not decorative: it marks the course
+   record and nothing else, so the eye learns one meaning for it. */
+
+:root {
+  --paper: #ffffff;
+  --wash: #f0f1f4;
+  --sunk: #e7e9ee;
+  --ink: #16171a;
+  --mute: #656a72;
+  --rule: #dee0e5;
+  --scarlet: #bb0000;
+  --open: #0e7a50;
+  --wait: #a65b00;
+  --closed: #767c85;
+
+  --display: "Bricolage Grotesque", "Trebuchet MS", sans-serif;
+  --body: "IBM Plex Sans", system-ui, sans-serif;
+  --data: "IBM Plex Mono", ui-monospace, monospace;
+
+  --gap: clamp(1rem, 2.5vw, 1.75rem);
+  --radius: 3px;
+}
+
+@media (prefers-color-scheme: dark) {
+  :root {
+    --paper: #131417;
+    --wash: #1a1c20;
+    --sunk: #232529;
+    --ink: #eceef2;
+    --mute: #9aa1ab;
+    --rule: #2c2f35;
+    --scarlet: #ff5c4d;
+    --open: #4cc38a;
+    --wait: #e0a44a;
+    --closed: #868d97;
+  }
+}
+
+* { box-sizing: border-box; }
+
+body {
+  margin: 0;
+  background: var(--wash);
+  color: var(--ink);
+  font-family: var(--body);
+  font-size: 16px;
+  line-height: 1.5;
+  -webkit-text-size-adjust: 100%;
+}
+
+:focus-visible {
+  outline: 2px solid var(--scarlet);
+  outline-offset: 2px;
+}
+
+.shell {
+  max-width: 62rem;
+  margin: 0 auto;
+  padding: var(--gap);
+}
+
+/* Masthead ---------------------------------------------------------------- */
+
+.masthead {
+  display: flex;
+  align-items: baseline;
+  gap: 0.75rem;
+  flex-wrap: wrap;
+  margin-bottom: 1.25rem;
+}
+
+.wordmark {
+  font-family: var(--display);
+  font-size: clamp(1.75rem, 5vw, 2.5rem);
+  font-weight: 800;
+  letter-spacing: -0.03em;
+  margin: 0;
+  line-height: 1;
+}
+
+.tagline {
+  margin: 0;
+  color: var(--mute);
+  font-size: 0.95rem;
+}
+
+/* Search ------------------------------------------------------------------ */
+
+.search {
+  display: grid;
+  grid-template-columns: 1fr auto auto;
+  gap: 0.5rem;
+  background: var(--paper);
+  border: 1px solid var(--rule);
+  border-radius: var(--radius);
+  padding: 0.5rem;
+  position: sticky;
+  top: 0.5rem;
+  z-index: 10;
+}
+
+.search input,
+.search select,
+.search button {
+  font: inherit;
+  border-radius: var(--radius);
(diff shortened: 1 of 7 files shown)


EnesYilmazcode (commented):
**Review: one blocking bug, plus three smaller ones.**

Catching the bad seat data and pulling it rather than shipping it was the right call, and the ranking work is well evidenced. But the acceptance criterion "searching a professor's last name returns their classes" does not actually hold, and the failure is silent.

## BLOCKING: instructor search returns nothing for common names

`app.js` only ever requests page 1. That is fine when relevance puts your match on page 1, and it does for the examples in the PR description. It does not hold generally:

```
q=Smith    totalItems= 674  totalPages=4   matching courses on page 1 = 0    beyond page 1 = 123
q=Bucci    totalItems=  17  totalPages=1   matching courses on page 1 = 9    beyond page 1 =   0
q=Zweben   totalItems=  11  totalPages=1   matching courses on page 1 = 3    beyond page 1 =   0
```

Searching `Smith` surfaces **zero** of that professor's 123 courses. Not "fewer", zero. The results list fills with unrelated courses whose text happens to match, and there is no indication anything was withheld, so the user concludes the professor teaches nothing this term.

The PR was tested with two uncommon surnames, both of which fit on one page. That is exactly the sample that hides this.

Fix by fetching several pages and merging before ranking. `mergeCourses()` already exists and already dedupes by class number, so it takes the extra pages without change. Paging being non-deterministic argues for this rather than against it: merging several pulls increases coverage rather than reducing it.

## 2. The result count reports a page, not a total

The status line reads `31 courses, 200 sections in Autumn 2026` while `totalItems` was 1281 across 7 pages. Those numbers describe what got rendered, phrased as though they describe the term. Once paging is fixed the number will be closer, but it should still not imply completeness it cannot promise.

## 3. Concurrent searches can render stale results

`runSearch()` has no request sequencing. Submit, then change the term and submit again, and whichever response lands last wins. That is not necessarily the newer one. Track a request id and drop responses that are not the newest.

## 4. Sort comparator can throw on a malformed record

```js
.sort((a, b) => b.score - a.score || a.entry.course.catalogNumber.localeCompare(...))
```

`catalogNumber` is read straight from the API with no guard. One record missing it takes down the whole result list, since the throw happens inside the sort. Everything else in this PR defends against missing fields, so this is inconsistent as much as it is risky.

Blocking on 1. Fixing 2 and 3 alongside it since they are all in `app.js`. 4 is a one-line guard.

EnesYilmazcode:
All four fixed.

**1. Multi-page search.** `searchAllPages()` fetches up to 5 pages and merges through the existing `mergeCourses()`. The blocking bug is gone:

```
q=Smith      pages=4/4  courses=256  matching instructor=110   (was 0)
q=Bucci      pages=1/1  courses=  9  matching instructor=  9
q=CSE 2221   pages=5/7  courses=102  top: CSE 2221
```

Checked that ranking still puts real matches on top rather than just returning more noise. Top 5 for `Smith`, with the name that matched:

```
ACCTMIS 2200   Marc Harris Smith
ACCTMIS 2300   Marc Harris Smith
GENED 1201     Jeremie S Smith
BUSFIN 3500    Ivan C. Smith III
EHE 1100       Casey Sara Smith
```

Cost is 5 requests instead of 1, issued in parallel. Measured end to end at **660ms**, which is fine for a search someone triggers deliberately. Individual page failures degrade to an empty page rather than failing the whole search.

**2. Count wording.** Now reads `Showing N courses, M sections in Autumn 2026.` and appends `Narrow the search to see more.` when `totalPages > pagesFetched`, so it never implies completeness it cannot promise.

**3. Request sequencing.** Each search takes an incrementing id and stale responses return before touching the DOM. Also stops a stale response from re-enabling the submit button mid-flight.

**4. Comparator.** Coerced through `String(... ?? "")`.

Merging.

PR #15, opened 2026-08-19
Group sections by instructor

### LINKED ISSUE #5: Group sections by instructor
This is the whole point of the project, and the thing OSU's own search does not do. Every other course search leads with the section number and buries the human. Finder leads with the instructor and nests their sections underneath, so picking a professor is the primary action rather than a detail you dig for.

**Scope**
- Within a course, group sections by instructor name
- One block per instructor, with their sections listed under them
- Sections with no listed instructor group under a clearly labelled "Instructor not listed" block rather than being dropped
- Handle a section taught by more than one person

**Acceptance criteria**
- A course with several professors renders one block per professor
- No section disappears from the page during grouping. Section count before and after grouping must match

EnesYilmazcode:
Closes #5

The structural inversion this project exists for. Every other course search leads with the section number and buries the human. Finder leads with the teacher.

```
CSE 2221  Software I: Software Components          4 credits · 22 sections

  Paolo Bucci               2 sections
  Steve Gomori              4 sections
  KT Vandergriff            4 sections
  Naomi Lynn Zweben         4 sections
  ...
```

## Design note
Sections are keyed by their **whole instructor set**, not by each individual. Filing a co-taught section under both names would double-count it and make a course look larger than it is. The acceptance criterion was explicit that no section may disappear, and the same reasoning says none may be duplicated.

`Instructor not listed` sorts last rather than being dropped, so a section with no assigned teacher is still visible and still countable.

## Verification
Ran the count-preservation check the issue asked for across six queries:

```
section-count preservation: 876 courses checked, 0 mismatches
```

Before and after grouping, every course has exactly the same number of sections.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -291,3 +291,27 @@ body {
   white-space: nowrap;
   border: 0;
 }
+
+/* Instructor blocks. The teacher is the heading and sections nest under them,
+   which is the inversion the whole project exists for. */
+
+.teacher { border-top: 1px solid var(--rule); }
+.course > .teacher:first-of-type { border-top: 0; }
+
+.teacher-name {
+  margin: 0;
+  padding: 0.7rem 1rem 0.2rem;
+  font-family: var(--display);
+  font-size: 1rem;
+  font-weight: 700;
+  letter-spacing: -0.01em;
+}
+
+.teacher-name.is-unlisted {
+  font-family: var(--body);
+  font-weight: 500;
+  color: var(--mute);
+}
+
+.teacher .sections > .section:first-child { border-top: 0; }
+.teacher .section { padding-top: 0.35rem; padding-bottom: 0.6rem; }
--- a/js/render.js
+++ b/js/render.js
@@ -9,6 +9,7 @@
 import { formatWhen, formatPlace, formatUnits, instructorsOf } from "./format.js";
 
 const COMPONENT_ORDER = ["Lecture", "Seminar", "Studio", "Laboratory", "Recitation"];
+const UNLISTED = "Instructor not listed";
 
 function el(tag, className, text) {
   const node = document.createElement(tag);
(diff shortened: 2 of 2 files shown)


EnesYilmazcode (commented):
**Review: one real bug, otherwise good.**

Keying by the whole instructor set is the right call and the reasoning is sound. The 876-course count-preservation check is the correct way to prove the acceptance criterion, and it passes.

## Bug: `surname()` sorts generational suffixes as the surname

`surname()` takes the last whitespace-separated token, which is wrong for any name carrying a suffix. These are real names from the Autumn 2026 data, not hypotheticals:

```
Ivan C. Smith III        -> sorts under: III
Marc Harris Smith Jr.    -> sorts under: Jr.
Paolo Bucci              -> sorts under: Bucci     ok
Naomi Lynn Zweben        -> sorts under: Zweben    ok
KT Vandergriff           -> sorts under: Vandergriff  ok
```

So professors with suffixes get filed under `I`, `J`, or `V` and land in an order no one scanning the list would predict. It is not a crash and nothing is lost, but instructor blocks are the primary navigation in this view, so their order is the feature.

Skip a trailing suffix token and take the one before it. Worth handling `Jr`, `Sr`, `II`, `III`, `IV`, `V`, with and without a trailing period.

## Not blocking, worth noting
- Dropping the instructor from each section row is right now that the block heading carries it, but it means a section row no longer stands alone if it is ever rendered outside a group. Nothing does that today.
- `.teacher .section:first-child { border-top: 0 }` plus `.course > .teacher:first-of-type` is doing the separator logic in two places. Fine, just fragile if the markup nests differently later.

Blocking on the suffix bug since it is a two-line fix.

EnesYilmazcode:
Fixed.

```
Ivan C. Smith III        -> Smith
Marc Harris Smith Jr.    -> Smith
Paolo Bucci              -> Bucci
Naomi Lynn Zweben        -> Zweben
KT Vandergriff           -> Vandergriff
Cher                     -> Cher        (mononym falls back to the whole name)
```

Re-ran count preservation after the change: 358 courses, 0 mismatches. The two styling nits are left as-is, noted rather than churned.

Merging.

PR #17, opened 2026-08-19
Filter search results instead of only ranking them

### LINKED ISSUE #16: Stop dumping the whole subject into results
Searching `CSE 2321` returns **103 courses and 1000 sections**. The course you asked for is first, which is what ranking bought us, but everything else is still on the page. You have to scroll past CSE 4193, 4998, 4998H, 4999, 6193, 6998, 8998 and more, most of which are independent study and thesis courses with 40 to 70 sections each that all read "Time to be announced / Location to be announced".

Ranking put the right answer on top. It never removed the wrong ones. Those are different jobs and only one got done.

## What to change

**1. When the query identifies one course, show that course.**
A query with a subject and a catalog number that exactly matches a course is unambiguous. Show it alone, and put everything else behind a "Show N related courses" disclosure rather than inline. Related courses are genuinely useful (CSE 2221 and CSE 5022 are the same class for different students) but they are not what was asked for.

**2. Demote independent study and research courses.**
Courses whose sections carry no meeting pattern and are components like Independent Study are almost never what a student is searching for. They are the bulk of the noise: CSE 4193 alone contributes 66 sections. Keep them reachable by name, drop them from generic results.

**3. Apply a relevance floor.**
Anything scoring far below the top result is noise. Cut it rather than ranking it 47th.

## What to add

Comparing instructors for one course is the actual job, and it is currently only implied by the layout. Make it explicit for the single-course case: every instructor teaching it, side by side, so you can pick. Ratings from #7 slot straight into this.

## Constraint
Keep the interface minimal. This should mean fewer things on screen, not a filter panel.

EnesYilmazcode:
Closes #16

Reported from real use: searching `CSE 2321` returns 103 courses and 1000 sections. Ranking put CSE 2321 first, then left CSE 4193, 4998, 4998H, 4999, 6193, 6998 and 8998 underneath it, most of them independent study listings with 40 to 70 sections that all read "Time to be announced".

Ordering noise correctly is not the same as removing it. This PR removes it.

## Effect

```
query        before                    after
CSE 2321     103 courses / 928 sec  ->   1 course /  11 sections
CSE 2221     101 courses / 916 sec  ->   1 course /  22 sections
MATH 1151    166 courses / 981 sec  ->   1 course /  74 sections
Bucci          9 courses /  17 sec  ->   4 courses / 12 sections
CSE          102 courses / 1000 sec ->  25 courses / 244 sections
```

## How
- **Exact subject and number answers with that course alone.** `CSE 2321` is unambiguous, so it gets a direct answer instead of a ranked list.
- **Independent study, research and thesis courses are demoted** out of generic results. Detected by having no meeting pattern on any section plus a non-teaching component, not by title alone.
- **Relevance floor** cuts the long tail for open-ended queries, capped at 25 courses.

## Nothing is thrown away
Everything demoted moves into a `Show N related courses` disclosure. Related courses matter: CSE 2221 and CSE 5022 are the same class for different students, and someone really might want CSE 4193.

Verified that no course can be lost:

```
q=CSE 2321   ranked=103  primary= 1 related=102  total=103  lost=0
q=CSE 2221   ranked=101  primary= 1 related=100  total=101  lost=0
q=Bucci      ranked=  9  primary= 4 related=  5  total=  9  lost=0
q=Smith      ranked=252  primary=25 related=227  total=252  lost=0
q=CSE        ranked=102  primary=25 related= 77  total=102  lost=0
q=MATH 1151  ranked=166  primary= 1 related=165  total=166  lost=0
```

`primary + related` always equals the full ranked set. I got this wrong on the first pass, where demoted courses vanished entirely, and the check above is what caught it.

## Note
This also delivers the compare-professors case without new UI. A single-course result is now every instructor teaching it, side by side, which is what the layout was already built for and what the noise was burying.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -315,3 +315,22 @@ body {
 
 .teacher .sections > .section:first-child { border-top: 0; }
 .teacher .section { padding-top: 0.35rem; padding-bottom: 0.6rem; }
+
+/* Related courses stay reachable without competing with the answer. */
+
+.related {
+  border: 1px solid var(--rule);
+  border-radius: var(--radius);
+  background: var(--paper);
+}
+
+.related > summary {
+  padding: 0.75rem 1rem;
+  cursor: pointer;
+  color: var(--mute);
+  font-size: 0.9rem;
+}
+
+.related > summary:hover { color: var(--ink); }
+.related[open] > summary { border-bottom: 1px solid var(--rule); }
+.related > .course { border: 0; border-radius: 0; border-top: 1px solid var(--rule); }
--- a/js/app.js
+++ b/js/app.js
@@ -1,5 +1,5 @@
 import { fetchTerms, defaultTerm, searchAllPages, ApiError } from "./api.js";
-import { rankCourses } from "./rank.js";
+import { filterCourses } from "./rank.js";
 import { renderResults } from "./render.js";
 
 const els = {
@@ -37,19 +37,16 @@ async function runSearch(q, term) {
   els.submit.disabled = true;
   setStatus("Searching...");
   try {
-    const { courses, totalPages, pagesFetched } = await searchAllPages({ q, term });
+    const { courses } = await searchAllPages({ q, term });
     if (requestId !== latestRequest) return; // a newer search already answered
-    const ranked = rankCourses(courses, q);
-    renderResults(els.results, ranked);
-    if (!ranked.length) {
+    const { primary, related } = filterCourses(courses, q);
+    renderResults(els.results, { primary, related });
+    if (!primary.length) {
       setStatus(`Nothing matched "${q}" in ${termName(term)}. Try a subject and number, like CSE 2221.`);
     } else {
-      const sections = ranked.reduce((n, e) => n + e.sections.length, 0);
-      const truncated = totalPages > pagesFetched;
-      setStatus(
-        `Showing ${ranked.length} course${ranked.length === 1 ? "" : "s"}, ${sections} sections in ${termName(term)}` +
-          (truncated ? ". Narrow the search to see more." : ".")
-      );
+      const sections = primary.reduce((n, e) => n + e.sections.length, 0);
+      const noun = primary.length === 1 ? "course" : "courses";
+      setStatus(`${primary.length} ${noun}, ${sections} sections in ${termName(term)}.`);
     }
   } catch (error) {
     if (requestId !== latestRequest) return;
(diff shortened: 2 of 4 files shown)


EnesYilmazcode (commented):
**Review: one real bug, one thing verified clean.**

The filtering is right and the primary-plus-related invariant is the correct thing to have tested, especially since it caught the first attempt losing courses outright.

## Bug: the collapsed disclosure builds all of its DOM eagerly

`renderResults()` calls `renderCourse()` for every related entry and appends it to the `<details>` before anything is shown. `<details>` hides its content, it does not defer building it. So every search constructs the entire result set as DOM and then hides roughly 90% of it:

```
q=MATH 1151   visible sections= 74   sections built inside the collapsed details= 926
q=CSE 2321    visible sections= 11   sections built inside the collapsed details= 917
```

For CSE 2321 that is 917 sections of DOM constructed so that 11 can be displayed. Each section builds five or six elements, so this is on the order of 5,000 nodes created and immediately hidden, per search, on a phone, during a registration window. It is exactly the cost this PR set out to remove, just moved somewhere less visible.

Build the related list on first open instead. `<details>` fires a `toggle` event, so render once on the first expand and keep it after.

## Verified clean: no false positives in demotion

The risk with `isIndividualStudy()` was over-matching, since the title regex includes `research` and `thesis`, and asynchronous online courses legitimately have no meeting days. Checked every CSE course for Autumn 2026:

```
demoted as individual study: 10
  CSE 4193 4998 4998H 4999 4999H 6193 6998 6999 8998 8999
  all component [Independent Study]
```

All ten are genuine, and every one was caught by the component check rather than the title regex. No scheduled course and no online lecture was demoted. The title fallback never fired, which is the outcome I wanted to confirm before trusting it.

## Nit, not blocking
`kept.includes(e)` inside a `filter` is quadratic. At 252 courses it is irrelevant, and it only runs once per search. Worth a `Set` if the result cap ever rises.

Blocking on the eager DOM build.

EnesYilmazcode:
Fixed. Related courses now build on the first `toggle` rather than up front.

Measured DOM node counts in a real browser, on load versus after expanding:

```
CSE 2321     nodes on load=  107   after expanding related= 6761
MATH 1151    nodes on load=  428   after expanding related= 6940
```

CSE 2321 drops from 6761 nodes to 107, a 98% reduction. The full tree still gets built if you expand it, which is the only point at which it is worth paying for.

The quadratic `includes` is left alone as noted, since it runs once per search over at most a few hundred items.

Merging.

PR #18, opened 2026-08-19
Snapshot RateMyProfessors ratings nightly

### LINKED ISSUE #6: Snapshot RateMyProfessors ratings nightly

EnesYilmazcode:
Closes #6

RateMyProfessors sends no CORS headers, so the browser can never call it. This
fetches the Ohio State roster in CI instead and commits it as a static file that
Pages serves like any other asset, so the site stays serverless.

## What is here

- `scripts/fetch-ratings.mjs`, Node 22, ESM, no dependencies, global fetch only.
  Pages the roster into `data/ratings.json`.
- `.github/workflows/ratings.yml`, nightly at 07:20 UTC plus manual dispatch,
  commits only when the file actually changed.
- `data/ratings.json`, the first snapshot, 7367 professors, 1.78 MB.

README already documented `node scripts/fetch-ratings.mjs`, and that command now
works as written, so the README is unchanged.

## Two things about the upstream API that shaped the design

**Unrated professors are dropped.** The roster is 8366 people but only 7367 have a
single rating. The rest come back with `avgRating` 0, `numRatings` 0 and no
would-take-again, so there is nothing to show next to a name.

**The upstream search index drifts mid-run.** Paging is by array offset (`after`
cursors are base64 of `arrayconnection:N`) and the index behind those offsets is
rebuilt periodically. A rebuild partway through a run shifts every later offset, so
a pass skips whatever moved across a page boundary. Every full run loses something:
measured raw edge counts were 8306, 8335, 8337 and 8366 of 8366.

The fix is page size 500, which takes the run to 17 requests instead of 42 and gives
the index far fewer chances to shift. Anything at or above 2000 is rejected upstream.

Drift does not threaten the output, because what it drops is the unrated tail that
gets filtered out anyway. That is a property of how the index is ordered, not luck.
Mapping every record to its index position on a complete 8366 pass:

```
decile 0: 24/837 unrated      decile 5:   1/837 unrated
decile 1:  2/837 unrated      decile 6:   0/837 unrated
decile 2:  1/836 unrated      decile 7:  10/836 unrated
decile 3:  1/837 unrated      decile 8: 124/837 unrated
decile 4:  0/836 unrated      decile 9: 836/836 unrated
```

The last rated professor sits at position 7405 of 8366. Everything past it is
unrated, and that tail is where new records get appended and where a rebuild
reshuffles. So drift lands on exactly the records the filter discards.

The size guard runs on what the fetch actually returned, so a genuine upstream
failure still fails the run loudly rather than writing a thin file.

## Verification

Measured on this branch against the live API, not assumed.

Full run:

```
page 17: 8366/8366
fetched 8366 of 8366, 7367 of them rated
Wrote 7367 professors to data/ratings.json (1784920 bytes)
```

7367 records, 1784920 bytes, about 21 seconds.

Determinism, eight full runs across both revisions of this branch:

```
43f1065c2d9d68cc94674840b137db65a8cc43040f67720c9c677f01446d2149  x8
```

Byte identical, confirmed with `cmp`, and the same hash as the committed
`data/ratings.json`. Several of those runs fetched only 8306 to 8337 raw edges and
still produced the identical file.

The actual acceptance criterion, run after committing the file:

```
$ node scripts/fetch-ratings.mjs
$ git status --porcelain -- data/ratings.json
(empty)
```

No change, so the workflow skips the commit.

Size guard, forced by pointing the script at a school id that returns nothing:

```
Refusing to write: got 0 rated professors, expected at least 5000.
Upstream reported resultCount 0. Leaving data/ratings.json untouched.
exit 1
```

The existing file kept its hash and no partial temp file was left behind. Writes go
to a temp file and get renamed, so a crash cannot leave a half written file.

Known professor resolves:

```json
{
  "legacyId": 196250,
  "firstName": "Paolo",
  "lastName": "Bucci",
  "department": "Computer Science",
  "avgRating": 3,
  "numRatings": 147,
  "avgDifficulty": 4.1,
  "wouldTakeAgainPercent": 40.2439
}
```

`.github/workflows/ratings.yml` parses under `yaml.safe_load`.

## Notes for review

- There is deliberately no fetch timestamp in the JSON. It would change every night
  and defeat the whole commit-only-when-changed check. The commit date already
  records when it ran.
- `-1` is what upstream uses for "no data" on the three score fields. It is
  normalized to `null` so it cannot render as a real score.
- Regional campuses (Marion, Lima, Newark, Mansfield, ATI) are separate schools
  upstream and are not included.
- An earlier revision merged each run against the previous snapshot. It was measured
  to have no effect on the output and it made upstream deletions permanent, so it is
  gone. See the comment below for the numbers.
- I have not enabled or triggered the workflow. It needs a run on `main` before the
  schedule means anything.


EnesYilmazcode (commented):
**Review: one incorrect guarantee, otherwise strong work.**

The drift discovery is the valuable part of this PR and it was found the right way, by running the thing three times instead of trusting that stable sorting implies stable output. Diagnosing it down to `arrayconnection:N` offsets shifting under a rebuilt index, and confirming zero shared records disagreed on values, is exactly the level of proof this needed.

## The stated guarantee about removals is not what the code does

The report says "fresh data always wins, so genuine removals still propagate". `mergeWithPrevious()` does not do that:

```js
for (const record of previous.professors ?? []) merged.set(record.legacyId, record);
const before = merged.size;
for (const record of fresh) merged.set(record.legacyId, record);
```

That is a union keyed by `legacyId`. Fresh values overwrite matching ids, which is why field updates propagate, but a professor **absent** from the fresh fetch is never removed. If someone is deleted upstream they stay in `data/ratings.json` forever. There is no age-out and no seen-this-run marker, so absence and drift are indistinguishable and both resolve to "keep".

## Which raises a sharper question about whether the merge earns its place

The justification for merging is drift. But the diagnosis says every record that differed between runs had `numRatings: 0`, and the last line of `mergeWithPrevious()` is:

```js
return [...merged.values()].filter((p) => p.numRatings > 0);
```

So the records the drift actually touched are the same records that get filtered out anyway. If that holds, the merge is not protecting against the measured problem, and its main practical effect is making removals impossible.

Please check directly: run a fetch with the merge disabled and diff the result against the merged output. If they are identical, drop the merge, since page size 500 plus the `MIN_EXPECTED` guard already covers the failure mode. If they differ, keep it but add a bounded age-out (drop a record not seen for N consecutive runs) and correct the claim in the PR body.

## Verified, no action needed
- **Transfer size is fine.** 1.78 MB raw worried me for a phone during registration, but it is **174 KB gzipped**, which GitHub Pages serves automatically, and it is cached after first load. Not worth trimming fields.
- **Workflow is correct.** `git status --porcelain` rather than `git diff` so an untracked first run counts as a change is the right call, and the `pull --rebase` before push covers schedule and dispatch overlapping. Concurrency group is a good touch.
- **Omitting a fetch timestamp was right** and for the stated reason. The commit date carries it.
- Guarding on the fresh fetch before the merge, so the merge cannot mask an upstream failure, is the correct ordering.

Blocking only on the merge question, since the PR currently documents behaviour the code does not have.

EnesYilmazcode:
You were right on both counts, and the sharper version of the question was the
correct one. Took path 2: `mergeWithPrevious` is deleted in c021208.

**The removal bug was real.** A deleted professor is never returned with
`numRatings: 0`, they are simply absent from the fetch, and a union keyed by
`legacyId` cannot tell absence from drift. Both resolved to "keep", so a deleted
record would have persisted forever. My claim that removals propagated was wrong.

**The merge was also inert.** Six full runs with the merge disabled, diffed against
the merged output:

```
43f1065c2d9d68cc94674840b137db65a8cc43040f67720c9c677f01446d2149  no-merge run 1
43f1065c2d9d68cc94674840b137db65a8cc43040f67720c9c677f01446d2149  no-merge run 2
43f1065c2d9d68cc94674840b137db65a8cc43040f67720c9c677f01446d2149  no-merge run 3
43f1065c2d9d68cc94674840b137db65a8cc43040f67720c9c677f01446d2149  no-merge run 4
43f1065c2d9d68cc94674840b137db65a8cc43040f67720c9c677f01446d2149  no-merge run 5
43f1065c2d9d68cc94674840b137db65a8cc43040f67720c9c677f01446d2149  no-merge run 6
43f1065c2d9d68cc94674840b137db65a8cc43040f67720c9c677f01446d2149  committed (merged)
```

All seven identical, `cmp` clean. Those runs were not drift free either. Raw edge
counts were 8337, 8335, 8337, 8306, 8335 and 8337 of 8366, so every one of them lost
between 29 and 60 records to drift and every one still produced exactly 7367 rated
records. The merge had nothing to carry forward. Its own log line said so during the
earlier verification and I did not read it closely enough: `0 carried forward`.

**Why it holds, so this does not rest on six lucky runs.** I mapped every record to
its position in the index on a complete 8366 pass. The last rated professor is at
position 7405, and everything past it is unrated (decile 9 is 836/836, decile 8 is
124/837). The tail is where new professors get appended and where a rebuild
reshuffles, so drift lands on exactly the records the `numRatings > 0` filter already
discards. It is structural, not coincidence.

One honest caveat. This is not a proof of impossibility. At the original PAGE_SIZE
200, over 42 requests, I did once see a rated record lost (7367 vs 7366). At 500 it
has not happened across nine full runs. If it ever does, the result is a one record
diff in one nightly commit that heals itself the next night, which is a much better
failure than a professor who can never leave the file.

Net effect: 23 lines and the `readFile` import gone, the removal bug gone, and the
output is unchanged. PAGE_SIZE 500 and `MIN_EXPECTED` carry the failure mode on their
own, as you said.

PR description updated so the documented behaviour matches the code. Also dropped the
transfer size note since you have verified it, and left the fields alone.

EnesYilmazcode (commented):
**Approving.** The follow-up did the right thing: it settled the question with evidence instead of argument, and then went past what I asked for.

Six no-merge runs producing byte-identical output to the merged file is the direct answer. What makes it convincing rather than lucky is that those runs were **not** drift-free (8337, 8335, 8337, 8306, 8335, 8337 of 8366 raw edges) and every one still yielded exactly 7367 rated records. The merge genuinely had nothing to carry.

Establishing the mechanism was the part I did not ask for and is the part worth keeping. The last rated professor sits at index 7405 and everything beyond it is unrated, so cursor drift lands structurally on the records `numRatings > 0` already discards. That turns "it did not happen six times" into "here is why it does not happen", which is a different quality of claim.

Also correct to keep the caveat rather than round it off: one rated record was lost once at PAGE_SIZE 200 across 42 requests. Not claiming impossibility, and noting the failure mode self-heals on the next nightly run, is the honest framing. A one-record diff that corrects itself is strictly better than a professor who can never leave the file.

Confirmed on the branch: `mergeWithPrevious` has zero remaining references, and the PR body no longer documents behaviour the code does not have.

Merging.

PR #19, opened 2026-08-19
Snapshot Barrett seat counts daily

### LINKED ISSUE #13: Get real per-section seat counts

EnesYilmazcode:
Closes #13

The OSU class API returns one enrollment figure per course and repeats it onto every section, so a section that has been full for weeks reads as open. Barrett's schedule has real per-section numbers and the enrollment limit the API never exposes, but `asc.ohio-state.edu` sends no CORS headers, so the fetch happens in Actions and the result is committed.

This is the data pipeline only. Nothing is wired into the UI.

## What is here

- `scripts/fetch-seats.mjs`, Node 22 ESM, no dependencies, global fetch. Reads the subject index, pulls each subject file for the term, parses the fixed-width layout, writes `data/seats.json` keyed by class number.
- `.github/workflows/seats.yml`, daily at 12:30 UTC plus manual dispatch, commits only when the file changed.
- `docs/barrett-schedule.md`, the verified format so nobody has to rediscover it.
- `data/seats.json`, the Autumn 2026 snapshot.

Concurrency is 5 with a 120 ms delay per worker and three retries with backoff. Output is written to a temp file and renamed.

## Measured run, all subjects, term 1268

```
term 1268: 337 subjects listed
term 1268: 241 subjects offered, 17680 sections, 0 residue failures, 1080 continuation rows, 0 collisions, 12.5s
wrote data/seats.json (305211 bytes, 298.1 KiB)
```

96 of the 337 subjects 404 for this term, which just means they are not offered.

Summer 2026 was parsed as a second check: 198 subjects, 4866 sections, 0 residue failures.

## The residue check

After slicing out every known field the parser blanks those columns and asserts nothing but whitespace is left. A failing line is counted, never silently dropped, and the job refuses to write if more than 0.5% of lines fail. That is what will catch a layout change rather than shipping wrong seat numbers. Right now the failure count is zero across 22546 section lines in two terms.

## Correctness against the rows in the issue

```
class   Barrett line                        parsed
5168    32/40                               [32, 40, 0]
4827    40/40   full                        [40, 40, 0]
4831    41/40  +1  over cap plus waitlist   [41, 40, 1]
```

Day columns, checked on a section with W and F set and one with T and R set:

```
5169  CSE 2221  B   W F   0800A   ->  days "WF"
5168  CSE 2221  L  T R    0800A   ->  days "TR"
8787  AVIATN 2101 B T R S 0730A   ->  days "TRSa"
```

## Determinism

Two full live runs, same bytes:

```
88685c90d9c8b7aef3f38ab655f7519cc52ff15cb85aba2f445e2c1b2cb6dd0b  run 1
88685c90d9c8b7aef3f38ab655f7519cc52ff15cb85aba2f445e2c1b2cb6dd0b  run 2
```

Keys are sorted numerically and the only date in the file is Barrett's own `updated:` header, so nothing wall-clock leaks in.

## Two corrections to the column map in the issue

The class number sits at columns 24 to 31, not 20 to 31. Columns 20 to 22 hold a separate campus code, `NWK` or `WST`, and a wider slice swallows it into the class number:

```
     CSE 2111       NWK   4816 L                 M W      0355P-0545  RS0206      35/35       A.Soliman
```

Weekdays run to column 55, not 53. Saturday is an uppercase `S` at 54 and Sunday a lowercase `s` at 55. AVIATN 2101 meets on Saturdays.

Two smaller things the fixed map does not survive. Room is left aligned and enrollment right aligned, and both a long room name and a 4-digit enrollment break the alignment:

```
 CIVILEN 4001             7739 B  ( 7592)         T R     0530P-0735  BO0410/420   79/88       A.Massari
 ANATOMY 2300            35179 L                 M W F    0800A       ONLINE    1045/1050+4   K.Stover, E.Tkacz (TA)
```

So columns 69 to 94 are matched as a region rather than sliced. And the autoenroll field can carry two parents, `(30558,30559)`, which runs to column 46.

## Shape of the output

```json
{
  "term": "1268",
  "termName": "Autumn 2026",
  "source": "https://www.asc.ohio-state.edu/barrett.3/schedule/",
  "sourceUpdated": "2026-08-18",
  "note": "Barrett refreshes once a day around 06:50 Eastern. A missing class number means unknown, not zero.",
  "fields": ["enrolled", "limit", "waitlist"],
  "sections": { "4827": [40, 40, 0], "4831": [41, 40, 1] }
}
```

Three-element arrays keep 17680 sections at 298 KiB. Passing several terms nests them under `terms` instead.

## Notes

- The default term is derived from the date, first three digits are year minus 1900 and the last is the season, so no annual edit is needed. The workflow can still override it on dispatch.
- The schedule is a nightly batch. Every subject file shares one `Last-Modified` and one `updated:` date, so the numbers are a daily figure and should be labelled as one.
- Barrett has fewer sections than the API, 484 for CSE against roughly 1000, so an absent class number has to render as unknown and never as zero.
- 12:30 UTC is after Barrett's 06:50 Eastern rebuild in both standard and daylight time.


EnesYilmazcode (commented):
**Approving.** Six corrections to the column map I supplied, each backed by the specific lines that break the naive slice. That is the outcome I wanted from saying "verify it yourself, do not trust it blindly", and it is worth listing what would have shipped broken:

- Class number at 20..31 swallows a campus code, so `NWK   4816` parses as the class number. Regional-campus sections would have been silently corrupted rather than missing, which is the worse failure.
- Saturday and Sunday live at columns 54 and 55, not inside 49..53. 124 lines use column 54.
- Room, enrollment and waitlist cannot be fixed-sliced at all: a long room name shifts the row and a four-digit enrollment eats the space before the waitlist (`1045/1050+4`). Matching 69..94 as a region is the right call.

Treating the trailer tables as a hard stop is the detail I would most likely have missed. Reading the independent-study grid as sections would have produced plausible-looking wrong numbers, and plausible-looking wrong numbers are exactly what this whole issue exists to prevent.

## Verified independently

- **The 96 missing subjects are real.** I spot-checked four and got 200s, which looked like it contradicted the report, so I tested all 337: **241 return 200, 96 return 404.** My sample had just landed inside the 241. The claim holds exactly.
- **Coverage against real API sections is 100%,** which is better than I expected when I wrote the issue warning that seats would be missing for many sections:
  ```
  CSE 2221         22/ 22      MATH 1151    74/ 74
  ENGLISH 1110.01  42/ 42      PSYCH 1100   23/ 23
  CSE 2321         11/ 11      TOTAL      172/172 = 100%
  ```
  The gap I worried about was independent-study sections, which #17 now filters out of results anyway.
- Verification rows match Barrett exactly: `5168 [32,40,0]`, `4827 [40,40,0]`, `4831 [41,40,1]`. The 4831 row is the one that matters, since the OSU API calls that section Open while it sits at 41/40 with someone waiting.

The `note` field carrying "a missing class number means unknown, not zero" into the data itself is good practice. That distinction is the difference between a useful number and a lie, and putting it next to the data means the UI cannot accidentally forget it.

## One thing the wiring must respect

`data/seats.json` declares `term: 1268` and holds one term only, while the UI has a term selector offering Spring 2026 and Summer 2026. Joining these seats onto a Spring 2026 section would show confidently wrong numbers, which is worse than showing none. The UI must compare `seats.term` against the selected term and show nothing when they differ. Filing that as an explicit requirement on the wiring issue rather than leaving it implied.

Merging.

PR #22, opened 2026-08-19
Show RateMyProfessors ratings on instructor blocks

### LINKED ISSUE #7: Show ratings on instructor blocks
Join the ratings snapshot onto the instructor blocks from #5.

The join is the hard part. OSU returns full legal names like `Diana Ikenberry Kline` while RateMyProfessors has `Diana Kline`, so exact matching will miss. Match on first token plus last token, lowercased and stripped of punctuation, and fall back to a link out when there is no confident match.

**Scope**
- Load `data/ratings.json` once and index it
- Show rating, number of ratings, and difficulty on each instructor block
- Never imply precision that is not there: a professor with 2 ratings must not look equivalent to one with 200
- Where there is no match, link to a RateMyProfessors search instead of showing a blank

**Acceptance criteria**
- `Paolo Bucci` resolves and shows his rating
- A professor absent from RateMyProfessors renders cleanly with a search link, not an error

EnesYilmazcode:
Closes #7

Joins the #18 snapshot onto the instructor headings from #5. This is the payoff for the whole project: search a course, see who teaches it, see how they are rated, in one view.

## The number that shaped the design

**Only 36% of instructors are on RateMyProfessors.** Measured across 757 instructors in 8 subjects for Autumn 2026. So the unmatched case is the *common* case, and it has to cost nothing visually.

The instructor name itself is the link. Matched, it goes to their RMP profile. Unmatched, it goes to an RMP search for them. Either way an unrated professor adds no extra marks to the page, which keeps the interface as quiet as it was.

## Matching took two passes

Exact first-and-last matching, after stripping middle names and generational suffixes, got 33%. Investigating the misses showed two different causes:

```
Steve Gomori             RMP has 1 Gomori: Stephen Gomori (35)     <- fixable
Timothy Ellis Carpenter  RMP has 3 Carpenters, incl Tim (37)       <- fixable
Umme Hafsa Billah        RMP has 0 Billah                          <- genuinely absent
Ramin Yarinezhad         RMP has 0 Yarinezhad                      <- genuinely absent
Jeff Eden                RMP has 0 Eden                            <- genuinely absent
```

OSU publishes legal names, students type everyday ones. A surname fallback accepting a first-name prefix or, when the surname is unique, a shared initial, fixes that class:

```
Steve Gomori             -> Stephen Gomori   3.7 (35)
Timothy Ellis Carpenter  -> Tim Carpenter    4.4 (37)
Diana Ikenberry Kline    -> Diana Kline      4.2 (32)
Dauntrica Marie Woods    -> no match          (6 different Woods, correctly refuses)
Umme Hafsa Billah        -> no match          (nobody by that surname)
```

That took it 33% to 36%. The remaining 64% are simply not on RMP, mostly lecturers and newer instructors, and no matching cleverness will conjure them.

A wrong rating is worse than no rating, so ambiguity always loses: 62 name keys have two or more professors sharing a first and last name, and those return nothing rather than a coin flip.

## Rating counts are never optional
The issue asked that 2 ratings not look like 200. The count always renders, and anything under 5 is dimmed. Displayed as `★ 2.7 (3)` rather than `★ 2.7 / 3`, because the slash reads as a score out of three when 3 is the number of people who rated them.

Also adds the live URL to the README and the repo homepage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 Find who teaches your classes before everyone else.
 
+**Live: https://enesyilmazcode.github.io/Finder/**
+
 Ohio State's own class search will tell you a course exists. It is far less
 helpful when the question you actually have is "which of these six sections
 should I take, and who is teaching them." Finder answers that one instead:
--- a/css/finder.css
+++ b/css/finder.css
@@ -365,3 +365,29 @@ body {
 .related > summary:hover { color: var(--ink); }
 .related[open] > summary { border-bottom: 1px solid var(--rule); }
 .related > .course { border: 0; border-radius: 0; border-top: 1px solid var(--rule); }
+
+/* Ratings. The name is the link, so an unrated instructor costs nothing. */
+
+.teacher-link { color: inherit; text-decoration-color: var(--rule); text-underline-offset: 3px; }
+.teacher-link:hover { text-decoration-color: var(--scarlet); }
+
+.score {
+  display: inline-flex;
+  align-items: baseline;
+  gap: 0.3rem;
+  margin-left: 0.5rem;
+  font-family: var(--data);
+  font-size: 0.8rem;
+  font-weight: 500;
+  color: var(--mute);
+  vertical-align: baseline;
+}
+
+.score::before { content: "★"; color: var(--scarlet); font-size: 0.75rem; }
+.score[data-thin="true"] { opacity: 0.65; }
+.score[data-thin="true"]::before { color: var(--mute); }
+/* Parenthesised, because "2.7 / 3" reads as a score out of three when 3 is
+   actually how many people rated them. */
+.score-count { font-size: 0.7rem; color: var(--mute); }
+.score-count::before { content: "("; }
+.score-count::after { content: ")"; }
--- a/js/app.js
+++ b/js/app.js
@@ -1,6 +1,7 @@
 import { fetchTerms, defaultTerm, searchAllPages, ApiError } from "./api.js";
 import { filterCourses } from "./rank.js";
 import { renderResults } from "./render.js";
+import { loadRatings } from "./ratings.js";
 
 const els = {
   form: document.querySelector("#search"),
@@ -72,6 +73,10 @@ function termName(code) {
 }
 
 async function init() {
+  // Fire and forget. Ratings are an enhancement, so a failed or slow load must
+  // never hold up or break a search.
+  loadRatings().catch((error) => console.warn("ratings unavailable", error));
+
   const params = new URLSearchParams(location.search);
   setBusy(false);
   setStatus("Loading terms...");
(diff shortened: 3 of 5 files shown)


EnesYilmazcode (commented):
**Review: one race condition, blocking.**

The 36% measurement is the right thing to have led with, and letting it drive the design rather than the other way round is why the interface stayed quiet. Refusing to guess on the 62 ambiguous name keys is also correct: on a page whose entire purpose is choosing a professor, a confidently wrong rating is worse than a blank.

## Bug: ratings can lose the race against the search and never appear

`loadRatings()` is deliberately not awaited:

```js
// js/app.js:78
loadRatings().catch((error) => console.warn("ratings unavailable", error));
```

`ratingFor()` reads a module-level `index` that stays `null` until that resolves. So if a search finishes first, every instructor renders unrated, and nothing re-renders when the ratings land. The user sees a permanently rating-free page until they search again.

This is not theoretical. Measured against the live host:

```
ratings.json (1.86 MB)  0.52s
one search page         0.64s
```

Close on a fast connection, which is exactly why it passed in local testing and in my screenshot. But those two numbers scale differently: the search response is small and mostly latency-bound, while ratings.json is 1.86 MB and bandwidth-bound. On a phone on campus wifi the search wins comfortably and ratings silently do not show. The target user is on a phone.

The fire-and-forget instinct was right, the placement was not. Ratings genuinely should not *block* anything, but render does depend on them, so the two should be awaited together:

```js
const [{ courses }] = await Promise.all([
  searchAllPages({ q, term }),
  loadRatings().catch(() => null),
]);
```

That costs `max(search, ratings)` rather than the sum, only on the first search since the index is cached after, and a ratings failure still degrades to an unrated page rather than a broken one.

## Verified, no change needed
- Transfer cost is fine. 1.86 MB is 174 KB gzipped, served compressed by Pages and cached after first load.
- `rel="noopener noreferrer"` is set on every outbound link.
- `★ 2.7 (3)` over `★ 2.7 / 3` is the right call. I only noticed the slash reading as a denominator after seeing it rendered, which is an argument for looking at the page rather than the diff.

Blocking on the race.

EnesYilmazcode:
Fixed. Ratings are now awaited in a `Promise.all` with the search rather than fired and forgotten, so rendering cannot start without them. Cost is the slower of the two, not the sum, and only on the first search.

Verified end to end against live data, CSE 2221 for Autumn 2026:

```
teacher blocks : 8
ratings shown  : 8      every instructor resolved
thin-evidence  : 3      dimmed, under 5 ratings
profile links  : 8
search links   : 0
```

All eight instructors on this course matched, which is well above the 36% average and reflects that intro CSE lecturers have been around long enough to accumulate ratings. The three dimmed ones are Alpay (3 ratings), Briggs (7) and Goetz (4), sitting next to Bucci at 147, which is exactly the distinction #7 asked for.

Merging.

PR #23, opened 2026-08-19
Show real seat counts from the Barrett snapshot

### LINKED ISSUE #21: Show seat counts from the Barrett snapshot
#19 landed `data/seats.json` with real per-section enrolled, limit and waitlist for all 17680 Autumn 2026 sections. Nothing reads it yet.

Wire it onto section rows so a student can see that the 9:10am section is full before they plan around it.

## Requirements

**Guard on the term.** This is the one that will silently produce wrong numbers if missed. `data/seats.json` carries a single term and declares it as `term: "1268"`, while the UI offers Spring 2026 and Summer 2026 in the selector. Joining 1268 seats onto a Spring 2026 section would show confidently wrong data, which is worse than showing none. Compare `seats.term` against the selected term and render nothing when they differ.

**Missing means unknown, not zero.** The data file says so in its own `note` field. A class number absent from the snapshot must render as no information at all, never as `0/0` or an empty bar.

**Say how fresh it is.** Barrett refreshes once daily around 06:50 Eastern, so these are day-old figures, not live seats. During a registration window that difference matters. The file carries `sourceUpdated`.

**Full is the signal that matters.** Someone scanning ten sections wants to know which ones they cannot get into. A section at or above its limit, or with anyone waitlisted, should be obvious without reading numbers.

## Scope
- Load `data/seats.json` once, alongside the ratings load
- Join by class number onto section rows
- Keep it minimal. This is a number and a state on an existing row, not a new component

EnesYilmazcode:
Closes #21

Wires #19's snapshot onto section rows. This closes the loop opened back in #13, where seat counts were pulled out of the UI because OSU's API reports them wrong.

## The result that justifies the whole detour

CSE 2221, Autumn 2026, rendered from real data:

```
sections   : 22
with seats : 22
FULL       : 18
waitlisted :  6
status     : 1 course, 22 sections in Autumn 2026. Seats as of Aug 18.
```

**18 of the 22 sections are full.** OSU's own API calls all 22 of them "Open" with "32 enrolled". Anyone planning around that would pick a section that filled weeks ago.

## Term guard

The requirement most likely to be skipped and most likely to produce confident lies. The snapshot holds one term (`1268`) while the selector offers three, so `seatsFor()` refuses unless `snapshot.term` matches the selected term. Verified by switching:

```
Autumn 2026 (1268)   22 sections   22 with seats   "Seats as of Aug 18."
Spring 2026 (1262)   26 sections    0 with seats   no date claim
```

Spring shows no seats rather than Autumn's seats, and the status line drops the freshness claim too, since claiming a date for data you are not showing is its own small lie.

## Other requirements
- **Missing means unknown.** A class number absent from the snapshot renders nothing. Never `0/0`, never an empty bar.
- **Freshness is stated.** Barrett refreshes once daily around 06:50 Eastern, so the status line says `Seats as of Aug 18` rather than implying live seats.
- **Full is scannable.** Full sections render in scarlet with the waitlist appended, so scanning ten sections for the one you can still get into does not require reading every number.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -391,3 +391,21 @@ body {
 .score-count { font-size: 0.7rem; color: var(--mute); }
 .score-count::before { content: "("; }
 .score-count::after { content: ")"; }
+
+/* Seat counts. Full is the state worth spotting while scanning ten sections. */
+
+.seats {
+  grid-column: 3;
+  font-family: var(--data);
+  font-size: 0.78rem;
+  color: var(--mute);
+  text-align: right;
+  white-space: nowrap;
+}
+
+.seats[data-state="full"] { color: var(--scarlet); font-weight: 500; }
+.waitlist { margin-left: 0.25rem; opacity: 0.85; }
+
+@media (max-width: 34rem) {
+  .seats { grid-column: 1 / -1; text-align: left; }
+}
--- a/css/finder.css
+++ b/css/finder.css
@@ -241,38 +241,6 @@ body {
 
 .section-who { color: var(--ink); }
 
-.chip {
-  font-family: var(--data);
-  font-size: 0.75rem;
-  text-transform: uppercase;
-  letter-spacing: 0.04em;
-  white-space: nowrap;
-  display: inline-flex;
-  align-items: center;
-  gap: 0.35rem;
-}
-
-.chip::before {
-  content: "";
-  width: 0.5rem;
-  height: 0.5rem;
-  border-radius: 50%;
-  background: currentColor;
-}
-
-.chip[data-kind="open"] { color: var(--open); }
-.chip[data-kind="wait"] { color: var(--wait); }
-.chip[data-kind="closed"] { color: var(--closed); }
-.chip[data-kind="unknown"] { color: var(--mute); }
-
-.seats {
-  grid-column: 3;
-  font-family: var(--data);
-  font-size: 0.78rem;
-  color: var(--mute);
-  text-align: right;
-}
-
 .component {
   font-family: var(--data);
   font-size: 0.72rem;
--- a/js/app.js
+++ b/js/app.js
@@ -2,6 +2,7 @@ import { fetchTerms, defaultTerm, searchAllPages, ApiError } from "./api.js";
 import { filterCourses } from "./rank.js";
 import { renderResults } from "./render.js";
 import { loadRatings } from "./ratings.js";
+import { loadSeats, seatsTerm, seatsUpdated } from "./seats.js";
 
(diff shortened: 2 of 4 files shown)


EnesYilmazcode (commented):
**Review: one cleanup needed, the substance is right.**

The term guard is implemented the way the issue asked and, more importantly, dropping the `Seats as of Aug 18` claim along with the numbers is a detail I did not specify. Stating a freshness date for data you are not displaying would have been its own small dishonesty.

Verified term threading independently, since a single missed call site would silently render seats without the guard. All six render paths pass it, including the `related` disclosure branch, which is the one that would have been easy to miss:

```
renderSection(section, term)                 render.js:104
renderCourse({course, sections}, term)       render.js:133
  sortSections -> renderSection(section, term)  render.js:158
renderResults(container, {...}, term)        render.js:167
  primary.map(entry => renderCourse(entry, term))  render.js:168
  related.map(entry => renderCourse(entry, term))  render.js:181
```

## Cleanup: dead CSS, including a duplicate `.seats` rule

`css/finder.css` carries two `.seats` blocks, at line 268 and line 397, both setting `grid-column: 3`. The first is a leftover from #4, when seat rendering was removed after the API data turned out to be wrong. The stylesheet kept the rules.

The same commit orphaned the `.chip` block, lines 244 to 266, roughly 23 lines including `::before` and four `data-kind` variants. `grep -rn "chip" js/ index.html` returns nothing, so none of it can ever match.

Two `.seats` rules is the part that actually bites. They currently agree on everything the second one does not override, so nothing renders wrong today, but the next person editing seat styling has a coin flip over which block wins. Delete the dead one and the `.chip` set.

## Measured, acceptable

First search now pulls both snapshots:

```
data/ratings.json   1858618 raw   180253 gzip  (176 KB)
data/seats.json      305212 raw    71265 gzip  ( 69 KB)
                                              --------
                                                245 KB
```

245 KB gzipped once, then cached. I checked what Pages actually sends: `Cache-Control: max-age=600` plus a weak ETag, and a conditional request returns `304` with zero bytes. So repeat visits within the day cost nothing. Fine as is, and worth revisiting only if a third snapshot appears.

Blocking only on the dead CSS, which is a deletion.

EnesYilmazcode:
Dead CSS deleted, 32 lines. Both the duplicate `.seats` block and the orphaned `.chip` set are gone, leaving exactly one `.seats` rule at line 365.

Re-verified rendering is unchanged afterwards:

```
sections  : 22
with seats: 22
FULL      : 18
ratings   :  8
```

Merging.

Issue #28, opened 2026-08-19
Add the section detail pane


EnesYilmazcode:
The right pane from wireframe A, which is the part that makes the extra width earn its place.

Selecting a section shows:
- Instructor name, with rating, difficulty and would-take-again as large figures
- Seats, as a bar plus exact enrolled, limit and waitlist, with the Barrett date
- Meeting days, time, room and instruction mode
- Other sections and courses the same instructor teaches this term, with their seats
- Course description and prerequisites

**Rules**
- Every figure must degrade honestly. Most instructors have no rating at all, most courses will have data missing somewhere, and a missing value shows as absent rather than zero or a dash pretending to be data.
- Nothing here needs a new data source. Ratings, seats and descriptions are already loaded.
- With no selection the pane should say something useful rather than sitting empty.

**Acceptance**
- Clicking any section fills the pane
- An unrated instructor renders cleanly with no empty rating block

Issue #27, opened 2026-08-19
Add the filter rail


EnesYilmazcode:
The left rail below the course pickers, from wireframe A.

**Filters**
- Days of the week, toggled
- Earliest start and latest end
- Minimum instructor rating
- Hide full sections
- Hide online
- Rated instructors only

**Rules**
- Filtering happens client-side on results already fetched. No extra API calls per keystroke.
- Never hide silently. Always state how many sections are filtered out and offer to show them, the same way #17 handled related courses.
- A filter that would empty the results should say so rather than showing a blank pane.
- Filters belong in the URL so a filtered view can be shared.

**Acceptance**
- Every filter updates results without a refetch
- Hidden counts are always visible

PR #30, opened 2026-08-19
Build the three-pane app shell

### LINKED ISSUE #25: Build the three-pane app shell
Wireframe A, merged in #24, is the direction. Build its shell.

The current layout puts everything in a narrow centre column and leaves both outer thirds of a desktop screen empty.

**Layout**
- Left rail, fixed width, for course selection and filters
- Centre, flexible, for results
- Right pane, fixed width, for the selected section's detail
- Slim top bar carrying the wordmark, the course picker and the term

**Responsive, and this is the part that matters**
Registration happens on a phone. Three panes cannot survive that, so the layout has to collapse rather than shrink. Below the breakpoint the rail becomes a collapsible sheet and the detail pane becomes a full-screen view pushed on selection, with a way back. No horizontal scrolling at 320px.

**Keep the existing visual language.** Same palette, same type, same restraint. This is a layout change, not a redesign. The wireframe is deliberately close to what already exists.

**Acceptance**
- Desktop uses the full width with no dead outer thirds
- Nothing overflows at 320, 360, 390, 414, 768, 1100, 1600
- Keyboard reachable throughout, focus visible

EnesYilmazcode:
Closes #25

Wireframe A's layout, wired to the live app.

## Responsive behaviour, measured

Collapse is a layout change, not a squeeze. Measured in sized iframes at eight widths:

```
  320px cols=1 rail=sheet detail=hidden filtersBtn=yes  overflow=no  spills=0
  360px cols=1 rail=sheet detail=hidden filtersBtn=yes  overflow=no  spills=0
  390px cols=1 rail=sheet detail=hidden filtersBtn=yes  overflow=no  spills=0
  414px cols=1 rail=sheet detail=hidden filtersBtn=yes  overflow=no  spills=0
  768px cols=1 rail=sheet detail=hidden filtersBtn=yes  overflow=no  spills=0
 1024px cols=1 rail=sheet detail=hidden filtersBtn=yes  overflow=no  spills=0
 1100px cols=3 rail=shown detail=shown  filtersBtn=no   overflow=no  spills=0
 1600px cols=3 rail=shown detail=shown  filtersBtn=no   overflow=no  spills=0
```

Zero overflow and zero spilling elements at every width, so #20's guarantee still holds.

## Interaction, measured

```
desktop after click: Section 5377 MoWeFr 3:00p-3:55p Lecture JR 239 50/50 +3
  selected rows   : 1
  results visible : true

mobile view attr  : detail
  results hidden  : true      detail shown : true      back btn shown : true
  focus moved to  : detail
  after back      : view=results  focus=results
  rail after toggle: open  aria-expanded=true
```

Focus management is the part worth reviewing. Collapsed, opening a detail hides the results entirely, so leaving focus on a section row inside a hidden pane would strand a keyboard user. Focus moves to the detail pane and returns to results on the way back.

Widening the window while a detail is open also resets the view, otherwise the results stay hidden behind a pane that is no longer covering anything.

## Scope
Detail content is a deliberate placeholder showing the row's own text. The real pane, with ratings, difficulty, other sections and the description, is #28. The rail is empty and says so, pointing at #27.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -61,25 +61,35 @@ body {
   :focus-visible { outline: 2px solid Highlight; }
 }
 
-.shell {
-  max-width: 62rem;
-  margin: 0 auto;
-  padding: var(--gap);
+/* Three-pane shell --------------------------------------------------------
+
+   Desktop is a real three-column application: filters, results, detail.
+   Below the breakpoint it becomes one column, because three panes squeezed
+   into a phone is worse than one pane at a time, and registration happens
+   on a phone. */
+
+.app {
+  display: grid;
+  grid-template-columns: 17rem minmax(0, 1fr) 23rem;
+  grid-template-rows: auto minmax(0, 1fr);
+  height: 100vh;
+  height: 100dvh;
 }
 
-/* Masthead ---------------------------------------------------------------- */
-
-.masthead {
+.topbar {
+  grid-column: 1 / -1;
   display: flex;
-  align-items: baseline;
+  align-items: center;
   gap: 0.75rem;
+  padding: 0.6rem var(--gap);
+  background: var(--paper);
+  border-bottom: 1px solid var(--rule);
   flex-wrap: wrap;
-  margin-bottom: 1.25rem;
 }
 
 .wordmark {
   font-family: var(--display);
-  font-size: clamp(1.75rem, 5vw, 2.5rem);
+  font-size: 1.4rem;
   font-weight: 800;
   letter-spacing: -0.03em;
   margin: 0;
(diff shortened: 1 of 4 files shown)


EnesYilmazcode (commented):
**Review: blocking. The new primary interaction is mouse-only.**

The layout and the collapse behaviour are right, and the focus management on the collapsed takeover is genuinely the hard part done correctly. But selecting a section is now the main thing you do in this app, and it cannot be done from a keyboard at all.

`.section` gets `cursor: pointer` and a delegated `click` handler, and nothing else. It is an `<li>` with no `tabindex`, no `role`, and no key handler:

```
sections rendered       : 8
sections focusable      : 0        <-
sections with role      : 0        <-
after .focus() active   : BODY     <- cannot even be focused programmatically
Enter changed detail    : false    <-
all focusable inside #results: a.teacher-link, a.teacher-link, a.teacher-link, summary
```

So a keyboard user can tab to the instructor links and the related-courses disclosure, and can never reach a single section. The detail pane is unreachable for them, which means the entire right third of the new layout is mouse-only.

This directly regresses #20, which established and measured "usable end to end with keyboard only". That PR checked focusability inside `#results` and correctly found nothing interactive there at the time. This PR made those rows interactive without making them focusable, so the earlier guarantee quietly stopped being true.

`cursor: pointer` is the tell. It is a visual claim that something is actionable, and whenever it appears on a non-interactive element the accessible role is usually missing too.

**Fix:** give each row a real interactive role, `tabindex="0"`, and Enter plus Space handling, and mark the selected row with `aria-current` so the state is exposed rather than being colour alone. The rows contain no nested links, so a button role is safe here.

Please re-run the focusable count and an Enter-key test after, and confirm the focus ring is visible on a row in both colour schemes.

## Not blocking
- `showDetail` is exported but nothing imports it. Either wire it from #28 or drop the export.
- `height: 100vh` followed by `height: 100dvh` is the right order for the fallback.

EnesYilmazcode:
Fixed.

```
sections rendered   : 8
sections focusable  : 8      (was 0)
sections with role  : 8      role=button
focus lands on row  : true   (was BODY)
Enter fills detail  : true   -> Section 5353 MoWeFr 4:10p-5:05p Lecture JR 239 50/50 +2
Space also works    : true
aria-current set    : 1      and stays exactly 1 after re-selecting
```

`Space` calls `preventDefault`, which matters here: without it Space scrolls the results pane instead of activating the row, so the keyboard path would have looked broken in a different way.

Selection now carries `aria-current="true"` rather than relying on the scarlet inset alone, so it is announced rather than merely visible.

On the focus ring: my first check tried `getComputedStyle(el, ':focus-visible')`, which returns empty because that reads pseudo-*elements*, not pseudo-*classes*. Not a failure, an invalid probe. Verified visually instead, and the scarlet ring renders on the focused row with the detail pane filled from the keypress.

Dropped the unused `showDetail` export as suggested.

Issue #29, opened 2026-08-19
Add the calendar view toggle


EnesYilmazcode:
Wireframe C's grid, as a view toggle in the centre column rather than its own layout. List stays the default.

**Explicitly out of scope: course planning.** No adding courses, no saved schedule, no conflict detection against other classes. Finder is a search tool. The calendar plots the sections of what you searched, nothing else. Wireframe C showed a "my schedule" rail and an "add to my schedule" button; both are cut.

**The hard part is overlapping sections.** Several instructors routinely teach the same course at the same hour. CSE 2321 has three sections at MoWeFr 3:00p, taught by Duong, Estill and Yarinezhad. Slicing that slot into three slivers makes it unreadable at any width.

Since the point of Finder is comparing instructors, a shared slot should render as **one block listing the instructors who teach it**, each with their rating and seats. That is more useful than three narrow cards and it degrades gracefully as the count grows.

**Acceptance**
- A course with three sections in one slot renders one readable block naming all three instructors
- Selecting a block fills the detail pane, same as the list view
- Sections with no meeting pattern (online, arranged) appear somewhere rather than being silently dropped

PR #31, opened 2026-08-19
Add the section detail pane

### LINKED ISSUE #28: Add the section detail pane

EnesYilmazcode:
Closes #28

The right pane from wireframe A, wired to real data. Nothing here fetches, since ratings, seats and descriptions are already loaded by the time you can click a section.

The shell's placeholder read text back out of the DOM. This replaces it with a class-number index of the real section and course objects built at render time.

## Decisions worth reviewing
- **Ratings only for a single named instructor.** Averaging two people is not a rating, so co-taught sections show none rather than a misleading blend.
- **Thin evidence is called out in words**, not just dimmed: under 5 ratings gets a line saying so.
- **"Also teaches" is scoped honestly.** It can only see courses the current search returned, so it is labelled "in these results" rather than implying it is the instructor's full load.
- Absent values render as absent. No zeros standing in for missing data.

## Three problems I can see in the rendered output, flagged in review below
I am not merging this as is. Screenshotting it surfaced issues the code reads fine for.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -498,11 +498,80 @@ body {
   margin: 0 0 0.6rem;
 }
 
-.detail-line {
-  margin: 0 0 0.4rem;
-  font-size: 0.9rem;
+/* Detail pane ------------------------------------------------------------- */
+
+.eyebrow {
+  font-family: var(--data);
+  font-size: 0.62rem;
+  letter-spacing: 0.14em;
+  text-transform: uppercase;
   color: var(--mute);
+  margin: 0 0 0.4rem;
 }
+
+.d-name {
+  font-family: var(--display);
+  font-weight: 700;
+  font-size: 1.25rem;
+  letter-spacing: -0.02em;
+  margin: 0 0 0.15rem;
+  overflow-wrap: break-word;
+}
+
+.d-name.is-none { color: var(--mute); font-family: var(--body); font-weight: 500; }
+
+.d-sub { margin: 0 0 1rem; color: var(--mute); font-size: 0.85rem; }
+
+.d-figs { display: flex; gap: 1.25rem; align-items: flex-end; margin-bottom: 0.9rem; flex-wrap: wrap; }
+
+.d-num {
+  font-family: var(--display);
+  font-weight: 800;
+  font-size: 1.9rem;
+  letter-spacing: -0.03em;
+  line-height: 1;
+}
+
+.d-num.is-rating { color: var(--scarlet); }
+.d-num.is-none { color: var(--mute); }
+
+.d-cap {
+  font-family: var(--data);
+  font-size: 0.62rem;
+  color: var(--mute);
+  margin-top: 0.2rem;
+}
+
+.d-block { margin-bottom: 1.1rem; }
+
+.d-row {
+  display: flex;
+  justify-content: space-between;
+  gap: 0.75rem;
+  font-size: 0.85rem;
+  padding: 0.32rem 0;
+  border-bottom: 1px solid var(--rule);
+}
+
+.d-row > span:first-child { color: var(--mute); }
+.d-val { font-family: var(--data); font-size: 0.8rem; text-align: right; }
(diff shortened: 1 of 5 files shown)


EnesYilmazcode (commented):
**Review: three problems, all visible only once rendered.** The code reads fine, which is the point. I found all three by screenshotting the pane, not by reading the diff.

## 1. "Also teaches" is full of exactly the noise #17 removed

Selecting Nickalaus Painter's section lists:

```
CSE 2331 · 5377    50/50
CSE 5331 · 5359     0/0
CSE 5331 · 5379     0/0
CSE 6331 · 5363    51/60
CSE 4193 · 9638       —
CSE 4998 · 9657       —
CSE 4998H · 9675      —
CSE 4999 · 9694       —
and 2 more
```

The last five are independent-study and thesis listings. Every professor is nominally attached to a dozen of them, so this section of the pane fills with rows nobody wants and pushes the real teaching load below the fold.

#17 already solved this with `isIndividualStudy()`, and this pane bypasses it by walking `currentEntries`, which includes `related`. Apply the same predicate.

## 2. A limit of zero renders as "0/0" in open-green

`CSE 5331 · 5359` shows `0/0` styled as if seats are available. It is meaningless: a capacity of zero cannot be open.

This is not an edge case. In the snapshot:

```
sections in snapshot : 17680
with limit of 0      : 1425  (8.1%)
  5359 -> [0,0,1]   5379 -> [0,0,1]
```

**8.1% of all sections**, and note `[0,0,1]`: zero capacity with someone on the waitlist, which is a real state Barrett records for unpublished or by-permission sections. `seatsFor()` currently returns these as normal seat data and computes `full` as `limit > 0 && enrolled >= limit`, so it correctly avoids calling them full, and then falls through to rendering them as open. Treat a limit of zero as unknown, the same as an absent class number.

## 3. The Barrett date is raw ISO in the pane and formatted in the status line

The pane says `As of 2026-08-18` while the status line above says `Seats as of Aug 18`. Same value, same screen, two formats. `formatDate()` already exists in `app.js`.

## Right calls, keep them
- Withholding a rating on co-taught sections rather than averaging two people.
- Labelling the block "in these results" instead of implying it is the instructor's full term load. That honesty is what made problem 1 visible.
- Stating thin evidence in words rather than only dimming it.

Blocking on 1 and 2. 3 is cosmetic but it is a one-line fix in the same file.

EnesYilmazcode:
All three fixed.

**1. Also teaches** now runs entries through `isIndividualStudy()` before listing them. Painter goes from 9 rows padded with CSE 4193, 4998, 4998H, 4999 and 2 more, down to the 4 sections he actually teaches.

**2. Zero limits** are treated as unknown in `seatsFor()` rather than rendered. `CSE 5331 · 5359` and `· 5379` now show `—` instead of a green `0/0`. The fix went in `seats.js` rather than the pane, since the same wrong number would otherwise reach the list view and the calendar in #29.

**3. Date** now reads `Aug 18` in both places, sharing `formatDate` rather than duplicating it.

Worth recording why all three got through: the code reads correctly for every one of them. `0/0` is a valid pair of numbers, an independent-study section is a real section, and an ISO date is a date. They only became obviously wrong when rendered next to each other. Screenshotting the pane was what caught them.

Merging.

PR #32, opened 2026-08-19
Add the filter rail

### LINKED ISSUE #27: Add the filter rail

EnesYilmazcode:
Closes #27

Fills the left rail. All filtering is client-side over results already fetched.

## Measured against live data, CSE 2221 Autumn 2026

```
baseline      22 sections
hide full      4 sections   "18 sections hidden by your filters. Show them anyway"
rating 4.0+    6 sections   "16 sections hidden by your filters. Show them anyway"
Mo + We        0 sections   "Your filters removed everything. Loosen one, or clear them."
url            ?q=CSE+2221&term=1268&day=monday&day=wednesday
after clear   22 sections   clear button hides itself again
```

The hide-full result lines up with the seat data: 18 of 22 CSE 2221 sections are full, so leaving 4 is exactly right.

## Contracts kept
- **Nothing hides silently.** Removed counts always show with a "Show them anyway" escape, same as #17 did for related courses.
- **Emptying the results says so**, and says it was the filters, rather than showing a blank pane that reads like a broken search.
- **Filters are in the URL**, so a filtered view reloads and shares.
- **No refetching.** Changing a filter repaints from the last search.

## Unknown data never fails a filter
A section with no meeting pattern cannot be judged on day or time. Failing it would make every online and arranged section vanish the moment someone touched a time control, which looks like data loss rather than filtering.

## One semantic choice worth a second opinion
Day filters mean **must meet on every day you pick**, which is why Mo + We returns nothing for CSE 2221: its lectures are TuTh and its labs WeFr, so no single section meets both. That is correct for "find me a MoWeFr class" and wrong for "I cannot do Fridays". The opposite need, excluding a day, is arguably more common during registration. Labelled explicitly in the rail so the behaviour is not a surprise, but this may want to flip to exclusion later.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -577,3 +577,94 @@ body {
   outline: 2px solid var(--scarlet);
   outline-offset: -2px;
 }
+
+/* Filter rail ------------------------------------------------------------- */
+
+.f-set { border: 0; margin: 0 0 1.25rem; padding: 0; }
+.f-set .eyebrow { padding: 0; }
+
+.f-days { display: flex; gap: 0.3rem; flex-wrap: wrap; }
+
+.f-day input { position: absolute; opacity: 0; width: 0; height: 0; }
+
+.f-day span {
+  display: block;
+  min-width: 2.1rem;
+  text-align: center;
+  padding: 0.35rem 0.3rem;
+  border: 1px solid var(--field);
+  border-radius: var(--radius);
+  font-family: var(--data);
+  font-size: 0.78rem;
+  color: var(--mute);
+  cursor: pointer;
+}
+
+.f-day input:checked + span {
+  background: var(--scarlet);
+  border-color: var(--scarlet);
+  color: var(--paper);
+  font-weight: 600;
+}
+
+.f-day input:focus-visible + span { outline: 2px solid var(--scarlet); outline-offset: 2px; }
+
+.f-hint { margin: 0.45rem 0 0; font-size: 0.72rem; color: var(--mute); }
+
+.f-label { display: block; font-size: 0.78rem; color: var(--mute); margin: 0.5rem 0 0.25rem; }
+
+.f-ctl {
+  width: 100%;
+  font: inherit;
+  font-size: 0.85rem;
+  padding: 0.4rem 0.5rem;
+  border: 1px solid var(--field);
+  border-radius: var(--radius);
+  background: var(--wash);
+  color: inherit;
+  cursor: pointer;
+}
+
+.f-check {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  font-size: 0.85rem;
+  color: var(--mute);
+  margin-top: 0.5rem;
+  cursor: pointer;
+}
+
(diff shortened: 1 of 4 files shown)


EnesYilmazcode (commented):
**Review: one contradiction on screen, one incomplete count.** Both only visible rendered.

## 1. "Your filters removed everything" sits directly above 46 surviving courses

Loading `?q=CSE 2221&term=1268&hideFull=1&rating=4` renders, top to bottom:

```
Your filters removed everything. Loosen one, or clear them.
  > Show 46 related courses
  22 sections and 1 course hidden by your filters. Show them anyway
```

Three statements, and they disagree. Something clearly was not removed, since 46 related courses are right there. The status is describing `primary` while the page is showing `primary` and `related` together.

The filtering itself is correct: 6 sections clear a 4.0 rating, all of them are full, so hide-full plus 4.0 legitimately leaves nothing. The wording is what is wrong. It should say no sections of *this course* match, which is both accurate and does not contradict the fold underneath it.

## 2. The hidden count only covers primary

`22 sections and 1 course hidden` counts what was removed from `primary`. `applyFilters` also runs over `related` and its removals are silently dropped on the floor:

```js
const related = showHidden ? lastResult.related : applyFilters(lastResult.related, filters).entries;
```

That discards the counts. So the number shown understates what the filters actually did, which breaks the same "never hide silently" contract this PR is otherwise careful about. Add the related totals in.

## Correct, keep it
- Unknown data never failing a filter is right, and it is the reason online sections do not vanish the instant someone touches a time control.
- The `Show them anyway` escape and the URL round-trip both work as measured.
- Hide-full leaving exactly 4 of 22 lines up with the seat snapshot independently, which is a good sign the filter is reading real data rather than something coincidental.

Blocking on both. They are the same defect really: the page reports on part of what it is showing.

EnesYilmazcode:
Both fixed. Same URL as before:

```
before
  status : Your filters removed everything. Loosen one, or clear them.
  note   : 22 sections and 1 course hidden by your filters.
  related: Show 46 related courses          <- contradicts the status

after
  status : No sections match your filters in Autumn 2026. Loosen one, or clear them.
  note   : 715 sections and 54 courses hidden by your filters. Show them anyway
  related: Show 45 related courses
```

The count was understating by a factor of thirty. `applyFilters` was already computing the related removals and the return value was being thrown away, so the note described a fraction of its own effect. That is exactly the silent hiding this PR set out to prevent, just one level up.

Merging.

PR #33, opened 2026-08-19
Add the calendar view toggle

### LINKED ISSUE #29: Add the calendar view toggle

EnesYilmazcode:
Closes #29

Wireframe C's grid as a view toggle in the centre column. List stays default.

## Course planning is deliberately absent
No saved schedule, no "add to my schedule", no conflict detection against other classes. The wireframe showed all three and they are cut. Finder searches; it does not plan. Adding a schedule means adding state, which means accounts, which is a different product.

## The overlap problem, which was the real design question

Several instructors routinely teach the same course at the same hour. CSE 2321 has three sections at MoWeFr 3:00p taught by Estill, Yarinezhad and Duong. Slicing a column three ways is unreadable, and since the point is comparing instructors, one block naming all three is clearer and more useful.

## Sizing was measured, not guessed, after guessing failed

First attempt used estimated constants. A DOM-presence check said everything was fine. A visibility check said otherwise:

```
instructor lines checked: 32, visually clipped: 9
  CLIPPED: "Luan Duong 4.9 44/40"  over by 22px    <- hidden entirely
  CLIPPED: "Ramin Yarinezhad 42/40" over by 1px
```

The third instructor in the exact slot this feature exists for was invisible, and my first probe passed because it counted DOM nodes rather than checking whether they were on screen.

So I measured the real rendered heights:

```
.cal-item   20.4px   (assumed 15)
.cal-time   14.4px   (assumed 13)
.cal-count  14.4px   (assumed 13)
a 3-section slot needs 96.4px, a 55-min slot gave 71.5px
```

Constants now carry the measured values with a note to re-measure if the type scale changes.

## After, across four courses

```
CSE 2321      lines= 32  most-shown-in-one-slot=3  truncated= 0  visually clipped=0
CSE 2221      lines= 44  most-shown-in-one-slot=2  truncated= 0  visually clipped=0
MATH 1151     lines= 81  most-shown-in-one-slot=3  truncated=14  visually clipped=0
ENGLISH 1110  lines=146  most-shown-in-one-slot=5  truncated=23  visually clipped=0
```

Zero clipping anywhere. Slots too deep to fit say `+N more, see list view` rather than dropping names, so the grid never quietly under-reports how many people teach a slot.

## Other details
- Only the hours actually used are drawn, so an afternoon course does not open with empty morning rows.
- Slot colour is worst-case: red when every section in it is full, amber when some are.
- Sections with no meeting pattern are listed below the grid rather than vanishing.
- Clicking any name fills the detail pane, same as the list.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -668,3 +668,113 @@ body {
   cursor: pointer;
   text-decoration: underline;
 }
+
+/* Calendar ---------------------------------------------------------------- */
+
+.view-bar { display: flex; margin-bottom: 0.8rem; }
+
+.view-toggle { display: inline-flex; border: 1px solid var(--field); border-radius: var(--radius); overflow: hidden; }
+
+.view-toggle button {
+  font: inherit;
+  font-size: 0.8rem;
+  padding: 0.32rem 0.8rem;
+  border: 0;
+  background: transparent;
+  color: var(--mute);
+  cursor: pointer;
+}
+
+.view-toggle button + button { border-left: 1px solid var(--field); }
+.view-toggle button.is-on { background: var(--scarlet); color: var(--paper); font-weight: 600; }
+
+.cal {
+  display: grid;
+  grid-template-columns: 3rem repeat(var(--cal-days), minmax(0, 1fr));
+  border: 1px solid var(--rule);
+  border-radius: var(--radius);
+  background: var(--paper);
+  overflow: hidden;
+}
+
+.cal-head {
+  font-family: var(--data);
+  font-size: 0.66rem;
+  letter-spacing: 0.08em;
+  text-transform: uppercase;
+  color: var(--mute);
+  text-align: center;
+  padding: 0.4rem 0;
+  border-bottom: 1px solid var(--rule);
+}
+
+.cal-col { border-left: 1px solid var(--rule); min-width: 0; }
+.cal-body { position: relative; }
+.cal-line { border-bottom: 1px solid var(--sunk); }
+
+.cal-gutter { min-width: 0; }
+
+.cal-hour {
+  position: relative;
+  border-bottom: 1px solid transparent;
+}
+
+.cal-hour span {
+  position: absolute;
+  top: -0.42rem;
+  right: 0.35rem;
+  font-family: var(--data);
+  font-size: 0.62rem;
+  color: var(--mute);
+}
+
+.cal-slot {
+  position: absolute;
+  left: 2px;
+  right: 2px;
+  border-radius: var(--radius);
+  border-left: 3px solid var(--open);
+  background: var(--wash);
+  padding: 0.2rem 0.3rem;
+  overflow: hidden;
+}
+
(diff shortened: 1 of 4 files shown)


EnesYilmazcode (commented):
**Review: one silent omission, blocking. Everything else measures clean.**

The measured-not-guessed sizing is the right correction, and catching your own DOM-presence check as insufficient is the part worth keeping. A node existing in the DOM and a node being visible are different claims, and only one of them was being tested.

Keyboard and selection verified independently:

```
cal item focusable : true
click fills detail : true
selected marked    : 1   aria-current=1
```

Using real `<button>` elements for slot entries rather than clickable divs means the keyboard path came for free, which is a nicer outcome than #30 had to be fixed into.

## Bug: switching to calendar silently drops related courses

```
LIST view
  related fold : Show 102 related courses
CALENDAR view
  related fold : ABSENT
  status says  : 1 course, 11 sections in Autumn 2026.
```

102 courses vanish on a view switch with nothing said about it. `paint()` passes only `primary` to `renderCalendar` and drops `related` on the floor.

Not plotting them is the right call. A week grid of 102 courses would be unreadable, and the grid is for comparing sections of the thing you searched. But this is the third time in this project the same defect has appeared: #17 hid related courses, #31 hid the instructor's real load behind independent-study rows, #32 undercounted filtered removals by a factor of thirty. Each time the fix was the same, say what is not being shown and offer a way to it.

Add a note under the grid: the count, and a control to switch back to list. It costs one line of UI and keeps the contract this project has otherwise held to.

## Also worth noting, not blocking
Slot colour being worst-case (red only when *every* section in the slot is full, amber when some are) is a good choice and not obvious. Worth keeping.

EnesYilmazcode:
Fixed.

```
calendar notes : "102 related courses are not on the grid. See them in list view"
after clicking : view is list
related fold   : Show 102 related courses
```

The escape actually switches view rather than just naming one, so it resolves the thing it mentions.

Merging.

Issue #26, opened 2026-08-19
Replace the search box with subject and number pickers


EnesYilmazcode:
Right now you have to know that Foundations II is CSE 2331. You should not have to.

Replace the single free-text field with two linked controls:
- **Subject**, a searchable list of all subject codes with their full names, so typing "computer" finds CSE
- **Number**, filtered to that subject, showing the catalog number alongside the course title, so you can pick 2331 by recognising "Foundations II"

Both should filter as you type and be usable from the keyboard alone.

**Keep free text working.** Typing `CSE 2331` or a professor's name straight in must still work, because it is faster for anyone who already knows what they want. The pickers are for when you do not.

**Where the data comes from.** Subjects and course numbers can be derived from search results, but doing that per keystroke is wasteful and the API's paged search is non-deterministic (see docs/osu-api.md), so a query could show a course one moment and not the next. Prefer building a course index in CI alongside the existing ratings and seats snapshots.

**Acceptance**
- Picking a subject then a number runs the search with no typing
- Typing a professor's name still works
- The number list shows titles, not bare numbers

PR #34, opened 2026-08-19
Build the course index for subject and number pickers

EnesYilmazcode:
Part of #26. This is the data half only. It adds no UI and touches nothing in `js/`, `css/` or `index.html`.

`scripts/fetch-courses.mjs` builds `data/courses.json`: for every searchable term, every subject with its full name, and every course under it with catalog number, title and credit range. `.github/workflows/courses.yml` rebuilds it weekly and commits only when it changed, same shape as the ratings and seats workflows.

## What it measured

Full run, all three searchable terms, on 2026-08-18.

| | Spring 2026 (1262) | Summer 2026 (1264) | Autumn 2026 (1268) |
|---|---|---|---|
| Subjects offered | 242 | 197 | 243 |
| Courses | 6163 | 1984 | 6072 |
| Sections behind them | 25274 | 15178 | 26293 |
| Passes needed | 3 | 3 | 3 |
| Courses the extra passes added | 0 | 0 | 0 |

Candidate subjects tried per term: 355. The ones that return nothing are dropped, which is why 355 becomes 243.

- Output: 771926 bytes, 181918 gzipped, so about 178 KiB over the wire
- Wall time: 2m10s for all three terms, about 2500 requests at 5 concurrent with a 120ms delay
- Autumn 2026 (1268) is covered, which was the minimum asked for

## Single pass versus reconciled

Reconciliation added nothing on the final runs, and that is because of a fix the measuring turned up rather than because the problem is not real.

In the API's default relevance order it is very real. Three whole-term builds of Autumn 2026 with one pass per subject, against the reconciled 6072:

```
single pass 1: 6068 courses, 4 missing   CHEM 6440, HISTORY 2350, 3002, 3351
single pass 2: 6069 courses, 3 missing   CSE 6521, CSE 6891, ECE 3551
single pass 3: 6064 courses, 8 missing   CHEM 6330, 6540, 6780, HISTORY 2025, ...
```

None of the three found a course the reconciled set lacked, so the union only grows toward the truth.

Then it turned out the cause is the ordering. Relevance order shuffles ties between pulls, a course drifts across a page boundary, and the walk misses it. Pass `sort=catalogNumber` and the pages stop moving. Eight identical pulls each:

| Subject | Pages | Relevance order | Catalog order | Union |
|---|---|---|---|---|
| PSYCH | 1 | 79 every time | 79 every time | 79 |
| CHEM | 4 | 57 to 58 | 58 every time | 58 |
| ECE | 7 | 72 to 73 | 73 every time | 73 |
| CSE | 6 | 102 to 104 | 104 every time | 104 |
| HISTORY | 4 | 107 to 111 | 111 every time | 111 |
| HTHRHSC | 3 | 61 to 65 | 65 every time | 65 |

Sorted, all 36 multi-page subjects returned the identical set four times running, and the sorted result equals the union of many unsorted pulls in every case. Only 36 of 243 subjects need a second page, so relevance order only ever hurt those.

The reconciliation loop stays. It stops after two consecutive passes that add nothing, capped at eight, and it says so if a subject is still growing at the cap. A stable order is something I measured, not something OSU promised. It also earns its keep as the thing that proved sorting works.

Worth recording that this was found the hard way: with the sort left off, three consecutive builds gave two identical files and a third that was 2 courses short, because two clean passes in a row is not a strong enough stop when the pages move.

## Determinism

Two consecutive full runs of the final script:

```
c6c4f5c993770285ac600c3f19098efc39464f0fb2c05ebd66e026ed119b9c56  data/courses.json
c6c4f5c993770285ac600c3f19098efc39464f0fb2c05ebd66e026ed119b9c56  data/courses.json
```

Both 771926 bytes. An earlier pair of runs on the same logic, before a wording change to one string in the file, also hashed identically to each other. There is no fetch timestamp in the file, on purpose, so a run with nothing new to say produces no commit.

## Spot checks

```
CSE 2331     "Foundations II: Data Structures and Algorithms"   3 credits
CSE 2221     "Software I: Software Components"                  4 credits
CSE 2231     "Software II: Software Development and Design"     4 credits
MATH 1151    "Calculus I"                                       5 credits
PSYCH 1100   "Introduction to Psychology"                       3 credits
ENGLISH 1110.01 "First-Year English Composition"                3 credits
```

Across all three terms: no blank titles, no blank subject names, no reversed credit ranges.

## Where the subject list comes from

Neither available source is complete on its own, so the script uses three and lets the API settle it.

The API has no subject endpoint (`/classes/subjects`, `/classes/subject`, `/classes/searchableSubjects` and `/classes/filters` all 404) and its own subject facet caps at ten entries, so it will not list them. Barrett's index is the only ready-made list, and it names 337 codes but is missing 18 that Autumn 2026 actually offers: AMINSTS, ASAMSTS, CIVICLL, CIVICTL, COMPEDU, CYBRSEC, ETHNSTD, MMT, NRO, RADONC, TLCTE, TLED, TLIELP, TLISTEM, TLLLL, TLTED, UKRAIN, UROLOGY. Sweeping the API's catalog-number buckets instead found 238 but missed five Barrett has and the API does offer.

So: Barrett as a seed, a sweep per term, plus every subject code the previous `courses.json` knew. A candidate that is not offered costs one request and is dropped, so the union costs little and nothing falls out of the file for a run.

## Shape of the file

```json
{"source":"...","fields":["catalogNumber","title","minUnits","maxUnits"],"note":"...",
 "terms":{"1268":{"term":"1268","name":"Autumn 2026","subjects":[
   {"code":"CSE","name":"Computer Science & Engineering",
    "courses":[["2221","Software I: Software Components",4,4]]}]}}}
```

Always keyed by term, even when the script is run for one, so the page reads one shape. Subjects sorted by code, courses sorted by catalog number with the leading digits compared as a number so 1151 lands before 2221 and 1151.01 sorts next to 1151.

One thing the UI half should know: 15 subjects have no `subjectDesc` upstream, so their name falls back to the code and the picker will show CYBRSEC as CYBRSEC. The subject facet title is the same string, so there is nothing better to use short of inventing names.

## Docs

`docs/osu-api.md` gets an "Enumerating the catalog" section with all of the above. Nothing already in it turned out to be wrong. Two things needed extending:

- The parameter table listed only `q`, `campus`, `term` and `p`. It now also lists `subject` and `sort`, which the response's own `filters` and `sort` arrays spell out.
- The paging note said to keep queries narrow enough to fit on one page. That is still the fallback, but it now points at `sort=catalogNumber` as the actual fix.

One small correction to the numbers already there: the note measures "1001 CSE sections, union 1064" from `q=CSE`, which also matches titles and instructor names and pulls in nine subjects. `subject=cse` reports 1064, exactly the union that note arrived at, so the subject filter's `totalItems` is the honest count.


EnesYilmazcode (commented):
**Approving.** The finding in here is worth more than the file it produced.

## Verified the headline claim independently

`sort=catalogNumber` making paging deterministic contradicts what `docs/osu-api.md` has said since #11, so I checked it rather than taking it:

```
subject=cse           totalItems=1064   pages=6
subject=CSE           totalItems=0                <- lowercase only, as reported

three identical pulls of subject=cse, term=1268:
  no sort              1062, 1064, 1064   intersection 1062  union 1064   NOT deterministic
  sort=catalogNumber   1064, 1064, 1064   intersection 1064  union 1064   IDENTICAL
```

Both claims hold exactly. This is a genuine correction to the project's own documentation: the "keep queries narrow enough to fit on one page" advice was a workaround, and there was a real fix available the whole time. Finding it by reading the response's own `filters` and `sort` arrays, rather than guessing parameter names, is the right instinct and something I should have done in #11.

The `subject=` parameter is the other half of it. I never found it, and it makes per-subject enumeration exact instead of approximate.

## Keeping the reconciliation loop was the right call

Reconciliation now adds zero, which makes it tempting to delete. The reasoning for keeping it is sound: a stable sort order is an observation about today's behaviour, not a guarantee, and the loop costs nothing when it converges on the first two passes. Also correct to measure the unsorted case anyway rather than skipping it once the fix was found. Losing CSE 6521, CSE 6891, ECE 3551, CHEM 6440 and HISTORY 2350 across single passes is a concrete demonstration that the problem was real.

Noting the honest detail that "two clean passes in a row" was not a strong enough stop condition while pages were moving, discovered by three builds producing two matching files and one short by two courses. That is the kind of thing usually quietly fixed rather than reported.

## Two things for the UI half

**1. Payload.** With this merged the browser pulls three snapshots:

```
ratings.json   176 KB gz
seats.json      69 KB gz
courses.json   177 KB gz
               -------
               422 KB gz on a cold load
```

Ratings and seats are needed to render any result. The course index is only needed once someone opens a picker. I will load it lazily in the UI half rather than at init, so first paint stays at 245 KB.

**2. The 15 subjects with no `subjectDesc`.** Flagging that CYBRSEC, TLED, MMT, NRO and the rest would render as "CYBRSEC CYBRSEC" was useful. The picker will show the code once when the name equals the code, rather than inventing names for them.

Merging.

PR #36, opened 2026-08-19
Add subject and course number pickers

### LINKED ISSUE #26: Replace the search box with subject and number pickers

EnesYilmazcode:
Closes #26

The UI half. #34 built the index this reads.

## Measured flow

```
on load        : ratings.json, seats.json                    <- courses.json absent
after focus    : ratings.json, seats.json, courses.json      <- loaded lazily
subjects       : 243
  first three  : ACADAFF — Academic Affairs | ACCAD — Adv Computing Cntr Arts & Design | ACCTMIS — Accounting and Management Info
  nameless one : CYBRSEC | MMT | TLED    (code once, not doubled)

picked CSE     : number field enabled, 104 options
  sample       : 1110 = Introduction to Computing Technology | 1111 = Introduction to Computer-Assisted Problem Solving
  search box   : "CSE"

picked 2331    : search box "CSE 2331"
  status       : 1 course, 8 sections in Autumn 2026. Seats as of Aug 18.
  url          : ?q=CSE+2331&term=1268
```

## Lazy loading was the point
With three snapshots the cold load would have been 422 KB gzipped. Ratings and seats are needed to render any result; the course index is only needed once someone opens a picker. Loading it on first focus keeps first paint at 245 KB and costs nothing to anyone who just types.

## The pickers build a query, they do not replace search
Choosing a subject and number writes into the existing search box and runs an ordinary search. So there is one search path, one URL format, and free text keeps working exactly as before. A professor's name still goes in the box.

## One bug worth recording
My first attempt read `index.terms[term]` as the subject array. It is an object with `term`, `name` and `subjects`, so the list came back empty and the picker silently offered nothing. I had guessed the shape from a summary instead of reading the file. Caught because the probe asserted on option counts rather than just that no error was thrown.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -778,3 +778,10 @@ body {
 .cal-empty { color: var(--mute); font-size: 0.9rem; }
 
 .cal-more { margin: 0; font-family: var(--data); font-size: 0.6rem; color: var(--scarlet); }
+
+/* Course pickers ---------------------------------------------------------- */
+
+.picker { margin-bottom: 1.25rem; padding-bottom: 1.1rem; border-bottom: 1px solid var(--rule); }
+.picker .f-ctl { cursor: text; }
+.picker .f-ctl:disabled { opacity: 0.5; cursor: default; }
+.picker .f-ctl::placeholder { color: var(--mute); }
--- a/index.html
+++ b/index.html
@@ -35,6 +35,20 @@
     </header>
 
     <aside class="rail" id="rail" aria-label="Filters">
+      <form id="picker" class="picker">
+        <p class="eyebrow">Find a course</p>
+        <label class="f-label" for="p-subject">Subject</label>
+        <input class="f-ctl" id="p-subject" list="subject-list" autocomplete="off"
+               placeholder="CSE, or computer" aria-describedby="p-hint">
+        <datalist id="subject-list"></datalist>
+
+        <label class="f-label" for="p-number">Course number</label>
+        <input class="f-ctl" id="p-number" list="number-list" autocomplete="off"
+               placeholder="2331, or foundations" disabled>
+        <datalist id="number-list"></datalist>
+        <p class="f-hint" id="p-hint">Pick a subject to browse its courses. You can still type a course or a professor in the search box.</p>
+      </form>
+
       <form id="filters">
         <fieldset class="f-set">
           <legend class="eyebrow">Meets on</legend>
--- a/js/app.js
+++ b/js/app.js
@@ -391,7 +391,7 @@ async function init() {
   writeFilters(params);
 
   for (const field of [els.subject, els.number]) {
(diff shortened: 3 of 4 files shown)


EnesYilmazcode (commented):
**Approving.** Two minor notes, neither blocking.

Lazy loading is the decision that matters and it is verified rather than asserted:

```
on load     : ratings.json, seats.json
after focus : ratings.json, seats.json, courses.json
```

422 KB cold load down to 245 KB, with the third snapshot arriving only for people who actually open a picker. Making the pickers write into the existing search box rather than becoming a second search path is the other good call: one query format, one URL, and free text keeps working unchanged.

Handling the 15 nameless subjects as "code once" rather than inventing names is right. `CYBRSEC`, `MMT` and `TLED` render correctly.

## 1. Picking a subject fires a search, then picking a number fires another

Typing `CSE` runs a full search for the whole subject before you have chosen a number, and choosing `2331` immediately runs a second. Each is five pages, roughly 660ms measured earlier, so the common flow costs two searches where one would do.

Searching on a bare subject is genuinely useful when that is all you want, so this is a trade rather than a bug. Worth revisiting if it feels laggy in use, most simply by waiting a beat after the subject before firing.

## 2. `{ once: false }` is the default

```js
field.addEventListener("focus", ensureCourses, { once: false });
```

Passing the default reads as though it were doing something. Drop the options object.

## On the structure bug you caught yourself
Reading `index.terms[term]` as the subject array when it is an object containing `subjects` is the kind of thing that fails silently: no error, no exception, just an empty picker. It surfaced because the probe asserted on option counts instead of only checking that nothing threw. Worth keeping that habit, since the same class of bug produced the empty rail and the clipped calendar lines earlier in this milestone.

EnesYilmazcode:
Dropped the redundant option. Leaving the double-search behaviour as noted, since searching a bare subject is genuinely useful on its own and the cost only shows up if it feels slow in real use.

Merging.

PR #43, opened 2026-08-19
Let day filters avoid days as well as require them

### LINKED ISSUE #41: Let day filters exclude days as well as require them
Day filters currently mean "the section must meet on every day you pick". Picking Mo and We returns nothing for CSE 2221, whose lectures meet TuTh and labs WeFr, because no single section meets both.

That is correct for "find me a MoWeFr class" and wrong for the more common registration need, which is "I cannot do Fridays". Both are legitimate, and the control currently expresses only one.

## Change

Let the day control express both: days required, and days to avoid.

Keep it small. This is one control in a rail, not a settings panel, and the interface should stay as quiet as it is now. Whatever interaction you choose, the meaning has to be obvious from looking at it, not from reading the hint underneath.

## Acceptance

- Avoiding Friday returns MoWe and TuTh sections and hides everything meeting on a Friday
- Requiring MoWeFr still works exactly as it does today
- The URL round-trips both modes, so a filtered view still shares and reloads
- The hidden-count note still reports accurately in both modes
- Unknown data still never fails a filter: a section with no meeting pattern must not vanish under either mode

EnesYilmazcode:
Closes #41

Day chips are now tri-state rather than on/off: **any → required → avoided → any**. One control, two meanings, no extra rows in the rail.

Avoided renders dashed and struck through, so it reads as "not this" without a legend.

## Measured, CSE 2221 Autumn 2026

```
baseline          22 sections
Fri required      11 sections   fr:require    the WeFr labs
Fri AVOIDED       11 sections   fr:avoid      the TuTh lectures
  url             ?q=CSE+2221&term=1268&noday=friday
  aria            "Friday: avoided"  aria-pressed=true
  sections still meeting Friday: 0
Fri back to any   22 sections
```

The two modes return complementary halves of the same 22 sections, which is the right shape, and nothing meeting Friday survives the avoid filter.

## Details
- **Buttons, not checkboxes**, since a checkbox cannot carry three states. Each announces its own state via `aria-label` ("Friday: avoided") rather than relying on colour and strikethrough, which a screen reader cannot see.
- **URL round-trips both modes**: `day=` for required, `noday=` for avoided.
- **Unknown data still never fails a filter.** A section with no meeting pattern has no days to judge, so neither rule applies and online sections do not vanish under either mode.
- Clear filters resets all chips to any, since `form.reset()` does not touch buttons.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -585,7 +585,13 @@ body {
 
 .f-days { display: flex; gap: 0.3rem; flex-wrap: wrap; }
 
-.f-day input { position: absolute; opacity: 0; width: 0; height: 0; }
+.f-day {
+  font: inherit;
+  padding: 0;
+  border: 0;
+  background: none;
+  cursor: pointer;
+}
 
 .f-day span {
   display: block;
@@ -597,17 +603,25 @@ body {
   font-family: var(--data);
   font-size: 0.78rem;
   color: var(--mute);
-  cursor: pointer;
 }
 
-.f-day input:checked + span {
+.f-day[data-state="require"] span {
   background: var(--scarlet);
   border-color: var(--scarlet);
   color: var(--paper);
   font-weight: 600;
 }
 
-.f-day input:focus-visible + span { outline: 2px solid var(--scarlet); outline-offset: 2px; }
+/* Avoided reads as struck out, which needs no legend to understand. */
+.f-day[data-state="avoid"] span {
+  border-style: dashed;
+  border-color: var(--mute);
+  color: var(--mute);
+  text-decoration: line-through;
+  text-decoration-thickness: 1px;
+}
+
+.f-day:focus-visible span { outline: 2px solid var(--scarlet); outline-offset: 2px; }
 
 .f-hint { margin: 0.45rem 0 0; font-size: 0.72rem; color: var(--mute); }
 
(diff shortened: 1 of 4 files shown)


EnesYilmazcode (commented):
**Approving.** Verified the parts I would have blocked someone else on.

```
chip focusable      : true
is a real button    : true
Enter changes state : require   sections 22 -> 11
focus retained      : true
clear resets chips  : true      sections restored 22
url after clear     : ?q=CSE+2221&term=1268     (day params removed)
```

Focus staying on the chip after activation matters here, because cycling through three states means pressing the same control repeatedly. Losing focus after the first press would make the avoid state unreachable by keyboard even though the button is focusable.

`form.reset()` not touching buttons was the trap in this change, and clearing them explicitly is handled. Worth noting that the chips stopped being form controls, so anything that assumed `FormData` carries the day state would silently read empty. `readFilters` reads them from the DOM instead.

The strikethrough plus dashed border carries the meaning without a legend, which was the requirement. Only the hint sentence explains the *interaction*, not the state, which is the right split.

## Non-blocking observation
Loading a URL with `?q=CSE 2221` does not populate the subject and number pickers, so the rail shows placeholders while the search box holds the query. Not a regression, and the pickers are a way to build a query rather than a mirror of it, but a shared link currently arrives with an empty-looking rail. Worth a small issue rather than a change here.

Merging.

PR #45, opened 2026-08-19
Give the landing screen something to do

### LINKED ISSUE #42: Give the empty state something to do
Landing on Finder shows an empty screen with "Type a course like CSE 2221, or a professor's name." The centre and detail panes are blank and the rail is a set of controls with nothing to act on. It is the first thing anyone sees and it does nothing.

There is already data in memory that would make this useful, and would show what the tool is for without explaining it.

## Ideas, pick what fits

- The subject picker already exists in the rail. Make it obviously the way in.
- Say something true from data already loaded: which term is being searched, when seats were last refreshed, how many instructors carry ratings.
- A worked example as a real link rather than a sentence, so one click shows the tool working on a real course.

## Constraints

- No marketing copy, no feature tour, no onboarding modal, no dismissable banner.
- Must not delay first paint, and must not pull the course index eagerly. #36 deliberately made that lazy to keep the cold load at 245 KB rather than 422 KB.
- Must collapse sensibly on a phone, where the rail is a sheet rather than a column.
- The empty state after a search that found nothing is a different case and is already handled. Do not change it.

EnesYilmazcode:
Closes #42

The first screen was an empty pane and a sentence telling you to type. It now shows the highest rated instructors at Ohio State, built entirely from snapshots already in memory.

```
requests on landing : ratings.json, seats.json
courses.json pulled : false                     <- #36's lazy loading intact

AUTUMN 2026 · 17,680 SECTIONS · SEATS AS OF AUG 18

Highest rated instructors at Ohio State
From 7,367 rated instructors, those with at least 50 ratings. Not all teach every term.

  Steve Yao              4.9    313 ratings · Fine Arts
  Amit Vutha             4.9    121 ratings · Mathematics
  Frederick Luis Aldama  4.9    105 ratings · English
  ...
```

Clicking a name runs a real search. Verified: clicking Steve Yao yields 24 courses, 34 sections.

## A claim I had to walk back
The heading first read "Best rated instructors **teaching this term**". That is not something this screen can know. The ratings snapshot covers every OSU instructor RateMyProfessors has, with no notion of who is teaching, and establishing it would mean one search per name on the landing screen.

It now says "Highest rated instructors at Ohio State" and the subtitle carries the caveat. Worth flagging because the false version looked better and read perfectly naturally.

## Constraints respected
- **No extra requests.** Ratings and seats are already in flight for the first search; this reuses them.
- **The course index stays lazy.** Verified `courses.json` is absent from the landing request list.
- **50-rating floor**, so the list means something. 332 instructors clear it, 948 clear 25, 95 clear 100.
- No marketing copy, no modal, no tour. A stat line, a list, and one worked example.

### DIFF
--- a/css/finder.css
+++ b/css/finder.css
@@ -799,3 +799,76 @@ body {
 .picker .f-ctl { cursor: text; }
 .picker .f-ctl:disabled { opacity: 0.5; cursor: default; }
 .picker .f-ctl::placeholder { color: var(--mute); }
+
+/* Landing screen ---------------------------------------------------------- */
+
+.welcome { max-width: 34rem; }
+
+.w-stats {
+  margin: 0 0 1.4rem;
+  font-family: var(--data);
+  font-size: 0.72rem;
+  letter-spacing: 0.06em;
+  text-transform: uppercase;
+  color: var(--mute);
+}
+
+.w-head {
+  font-family: var(--display);
+  font-weight: 700;
+  font-size: 1.15rem;
+  letter-spacing: -0.02em;
+  margin: 0 0 0.25rem;
+}
+
+.w-sub { margin: 0 0 1rem; color: var(--mute); font-size: 0.85rem; }
+
+.w-list { list-style: none; margin: 0; padding: 0; }
+
+.w-list li {
+  display: grid;
+  grid-template-columns: minmax(0, 1fr) auto;
+  gap: 0 0.75rem;
+  align-items: baseline;
+  padding: 0.55rem 0;
+  border-bottom: 1px solid var(--rule);
+}
+
+.w-name {
+  font: inherit;
+  font-weight: 600;
+  text-align: left;
+  background: none;
+  border: 0;
+  padding: 0;
+  color: inherit;
+  cursor: pointer;
+}
+
+.w-name:hover { text-decoration: underline; text-decoration-color: var(--scarlet); }
+
+.w-score {
+  font-family: var(--data);
+  font-size: 0.95rem;
+  color: var(--scarlet);
+}
+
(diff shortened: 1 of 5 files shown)


EnesYilmazcode (commented):
**Approving.** Verified the constraints the issue set, plus the two cases the PR did not cover.

```
with query, welcome hidden : true
after clearing query       : welcome hidden=false, names=8    <- returns correctly

at 360px: scrollW=360 clientW=360 overflow=no spills=0
  welcome visible : true      names listed : 8
  name focusable  : true      tag=BUTTON
```

Coming back cleanly when a search is cleared is the case most likely to have been missed, since it is easy to hide a landing screen once and never restore it. Names being real buttons means the keyboard path came free rather than needing retrofitting, which is the same thing that went right in #33 and wrong in #30.

## The walked-back claim is the important part of this PR

"Best rated instructors teaching this term" read perfectly naturally and was false. The ratings snapshot has no notion of who is teaching, and the screen had no way to know. Catching it required asking what the data can actually support rather than what the sentence sounded like.

Worth noting the pattern: this is the same failure as `0/0` rendering as open in #31 and the API's own per-section enrollment. In every case a plausible-looking number or sentence was presented as knowledge the system did not have. That is now three instances, and the fix has always been to narrow the claim rather than improve the guess.

The replacement is honest about both bounds: a 50-rating floor with the population it was drawn from, and an explicit "not all teach every term".

## Verified constraints
- `courses.json` still absent from the landing request list, so #36's lazy loading is intact
- No extra requests: reuses ratings and seats, which were already in flight
- No modal, no tour, no dismissable banner

Merging.

PR #46, opened 2026-08-19
Bring the README up to date

### LINKED ISSUE #40: Bring the README up to date
The README describes a product that no longer exists. It documents a single search box over a centred column, and says nothing about the three-pane layout, the filter rail, the calendar view, the detail pane, the course pickers, or two of the three nightly snapshots.

Anyone arriving at the repo builds a wrong mental model, and the architecture section understates the part that is actually interesting.

## Update it to describe what is there now

- What Finder does, in terms someone deciding whether to use it would recognise
- The three-pane layout and the list and calendar views
- All three snapshots, what each is for, and which load eagerly versus lazily and why
- How to run it locally, and how to run each pipeline by hand
- A short architecture diagram that reflects reality

## State the measured facts plainly

These are the things that make the project worth reading about, and all of them are already documented in the repo:

- OSU's own API reports enrollment that is wrong per section. It repeats one figure across every section of a course, calling full sections open. Barrett is snapshotted because of this.
- Only about 36% of instructors appear on RateMyProfessors, so the unrated case is the common one.
- Seats are a daily snapshot taken around 06:50 Eastern, not live.
- The site is static, with no server and no database. That is deliberate: the tool it replaces went down when its database container stopped resolving.

## Keep it honest

The interesting thing here is what had to be worked around, not a feature list. No marketing language, and no adjectives doing work that facts should be doing.

## Constraints

- `README.md` only. Do not touch `docs/osu-api.md`, another change is editing that file.
- Do not restructure the repo or move files.

EnesYilmazcode:
Closes #40

The README documented a single search box over a centred column and one nightly
snapshot. It now describes the three-pane layout, the filter rail and pickers,
the list and calendar views, the detail pane, and all three pipelines, with the
measured facts that explain why each one exists.

`README.md` is the only file touched.

## What is in it

- What Finder does, and the layout, as an ASCII sketch of the three panes.
- What each data source is, how fresh it is, and who fetches it. Sections and
  instructors come live from the browser; seats, ratings and the course index
  are committed snapshots.
- Why the API cannot supply seats: it repeats one `enrollmentTotal` across every
  section of a course and exposes no per-section limit, so it calls all 22 CSE
  2221 sections open at 32 enrolled when 18 of them are full.
- 36% RateMyProfessors coverage, and the name matching that took it from 33%.
- Which snapshots load eagerly and which lazily, and what that costs.
- A pipeline table with schedules, and the commands to run each one by hand.
- A repo layout map.

## Numbers I re-measured rather than copied

| Claim | Status |
|---|---|
| 36% of instructors on RMP, 757 instructors, 8 subjects | as stated in #22, kept |
| 18 of 22 CSE 2221 sections full | re-checked against the live API and `data/seats.json` today, still 18 |
| Seats are a daily batch around 06:50 Eastern | matches `docs/barrett-schedule.md` and the 12:30 UTC cron, kept |
| 245 KB eager, 422 KB with the course index | **423 KB now**, see below |
| Barrett has 484 CSE sections | 484, see below |
| 17680 section lines, zero residue, term 1268 | matches the committed snapshot, kept |

**The 422 KB figure has drifted to 423.** Those numbers are the gzipped weight of
the snapshots, not of the whole page. Measured on the committed files today at
`gzip -9`:

```
data/ratings.json   180253 B   176.0 KiB   eager
data/seats.json      71265 B    69.6 KiB   eager
                              -----------
                                245.6 KiB  cold load
data/courses.json   181922 B   177.7 KiB   lazy, on first picker focus
                              -----------
                                423.3 KiB  if the index were eager too
```

So the README says "about 245 KB" and "about 423 KB". The snapshots have grown a
little since #36 measured 422.

## Two things I found that contradict what is in the repo

1. **Not all three snapshots are nightly.** Ratings are daily at 07:20 UTC and
   seats are daily at 12:30 UTC, but the course index is **weekly**, Mondays at
   08:40 UTC, and `courses.yml` says why: it is the catalog rather than
   enrollment, and a full build is about 2500 requests. The README says weekly.

2. **`docs/osu-api.md` and `docs/barrett-schedule.md` disagree on the CSE section
   count.** The comparison table in `osu-api.md` says Barrett returns 468 CSE
   sections; `barrett-schedule.md` says 484. I parsed the live file for term
   1268 and counted **484** section lines, so `osu-api.md` is the stale one. I
   have not touched that file, since another change is editing it. The README
   uses 484, against the 1064 the API reports for `subject=cse`.

## Things worth fixing outside this PR

- **`seats.yml` cannot recover from a race the other two workflows handle.**
  `ratings.yml` and `courses.yml` both `git pull --rebase origin
  ${{ github.ref_name }}` before pushing, and push to `HEAD:${{ github.ref_name }}`.
  `seats.yml` does neither, it just runs `git push`. A scheduled seats run that
  overlaps a manual dispatch, or any commit landing on `main` during its 12
  second fetch, fails the push instead of rebasing over it.

- **The wireframes are stale in a way that reads as current.** `wireframes/README.md`
  opens with "the current single-column layout wastes the left and right thirds
  of a desktop screen", which was true when #24 landed and stopped being true
  when #30 shipped the three-pane shell. A line saying which one won would stop
  it reading as an open question.

## Style

No em dashes, no marketing language, and the MIT reference and the "not
affiliated with OSU" line are both still there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ Find who teaches your classes before everyone else.
 Ohio State's own class search will tell you a course exists. It is far less
 helpful when the question you actually have is "which of these six sections
 should I take, and who is teaching them." Finder answers that one instead:
-search a course, and get its sections grouped by **instructor**, with live seat
-status and RateMyProfessors ratings next to each name.
+search a course, and get its sections grouped by **instructor**, with per-section
+seat counts and RateMyProfessors ratings next to each name.
 
 ## Why this exists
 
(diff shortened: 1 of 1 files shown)


EnesYilmazcode (commented):
**Approving.** The three findings outside the README are worth more than the README rewrite, and all three check out.

## 1. The 468 is mine and it is wrong

Verified against the live Barrett file for term 1268, counting section lines with a correct parse:

```
Barrett CSE section lines, term 1268 : 484
distinct class numbers               : 484
present in data/seats.json           : 484
```

`docs/osu-api.md:292` says 468. `docs/barrett-schedule.md:125` says 484. The doc that is wrong is mine, from #11.

The cause closes a loop. My original count used a naive fixed-column slice at 20..30, which swallows the campus code on regional-campus rows so `NWK   4816` never parses as a class number. That is precisely the first of the six column-map corrections #19 reported after being told not to trust my map. So the stale figure was produced by the exact bug that PR later fixed, and it sat in the docs for another eight PRs.

Worth noting my own first attempt at re-counting it just now made the same mistake again and returned 0, because I hand-rolled the slice instead of reusing the parser that already knows better.

## 2. The seats workflow race is real

```
ratings.yml  pull --rebase: yes
seats.yml    pull --rebase: NO
courses.yml  pull --rebase: yes
```

`seats.yml` pushes directly, so any commit landing on `main` during its fetch fails the push instead of rebasing. The other two handle it. One-line gap, confirmed.

## 3. The wireframes README is stale
It still says the current layout wastes the outer thirds of the screen, which stopped being true when #30 shipped.

## On the corrections to my own prompt
"Three nightly snapshots" was wrong: the course index is weekly, on Mondays, and `courses.yml` says why. 422 KB was 423.3 KB measured at `gzip -9`. Both are small, and correcting them rather than repeating what I handed you is the behaviour I want. Noting the figures are snapshot weight rather than whole-page weight is a distinction I would have let slide.

All three follow-ups are being handled: 1 and 2 by the agents that own those files, 3 by me.

Merging.

PR #48, opened 2026-08-19
Snapshot seats for every searchable term

### LINKED ISSUE #39: Snapshot seats for every searchable term
`data/seats.json` holds one term. The term selector offers three.

#23 handled this honestly: `seatsFor()` refuses to return anything unless the snapshot's term matches the selected term, so choosing Spring 2026 shows no seat data at all rather than showing Autumn's numbers. That was the right call, and it leaves two of the three terms with no seats and a status line that quietly drops its freshness note.

Barrett publishes every term, so this is a limit of our pipeline rather than of the data.

## Change

- `scripts/fetch-seats.mjs` fetches every term the API reports as searchable, not only the current one
- `data/seats.json` keys sections by term
- `js/seats.js` reads the right term rather than guarding against the wrong one, and still returns null for a term genuinely absent from the snapshot

## Watch the size

One term is 305 KB raw and 69 KB gzipped. Three will be roughly triple, landing near the ratings snapshot. Seats load eagerly, unlike the course index, so this directly affects cold load. Measure it and report the number. If it is uncomfortable, say so rather than shipping a quiet regression in load time.

## Keep

- A zero limit still means unknown, not empty
- A missing class number still means unknown, never zero
- Deterministic byte-identical output for unchanged input
- The residue check that fails the job rather than shipping bad numbers

## Acceptance

- Selecting Spring 2026 or Summer 2026 shows real per-section seat counts
- A term with no Barrett data still degrades to no seats rather than wrong ones
- Report measured raw and gzipped sizes before and after

EnesYilmazcode:
Closes #39

`data/seats.json` held one term while the term selector offers three, so `seatsFor()` correctly refused to answer for the other two and Spring and Summer showed no seat data at all. Barrett publishes every term, so this was our pipeline, not the data.

`scripts/fetch-seats.mjs` now reads `searchableTermsV2` for the term list instead of deriving one from today's date, and snapshots each term off the same Barrett subject index. The parser, the column map and the residue check are untouched.

The snapshot is **one file per term** plus a small index, and `js/seats.js` fetches a term when it is asked for. Three terms in one eagerly loaded file was 154 KiB gzipped before anything renders, to carry two terms most visitors never select. Split, cold load is back to where it was.

```
data/seats.json          the index, 639 B, lists which terms exist
data/seats-1262.json     Spring 2026
data/seats-1264.json     Summer 2026
data/seats-1268.json     Autumn 2026
```

## Measured

Live runs, Node 22, 2026-08-18.

| term | subjects offered | sections parsed | residue failures | continuation rows | collisions | seconds |
|---|---|---|---|---|---|---|
| 1262 Spring 2026 | 241 | 17483 | 0 | 1142 | 0 | 12.5 |
| 1264 Summer 2026 | 198 | 4866 | 0 | 292 | 0 | 11.2 |
| 1268 Autumn 2026 | 241 | 17680 | 0 | 1080 | 0 | 11.6 |

34 to 36 seconds wall for all three, 1013 requests, politeness unchanged: 5 in flight, 120 ms between requests per worker, one shared subject index instead of one per term. A single term is 11 seconds and about 340 requests.

Autumn parses identically to the file this replaces: 17680 rows, 0 differing, same key order. The three rows from `docs/barrett-schedule.md` still hold, class 4827 is 40/40, 4831 is 41/40 with waitlist 1, 5168 is 32/40.

## Size

| | raw | gzipped |
|---|---|---|
| index | 639 B | 361 B |
| Spring 2026 | 300,887 B | 68,730 B (67.1 KiB) |
| Summer 2026 | 80,626 B | 17,555 B (17.1 KiB) |
| Autumn 2026 | 305,000 B | 70,990 B (69.3 KiB) |

**Cold load, index plus the term on screen: 71,351 B gzipped.** Before this PR, one term alone was 71,265 B. That is +86 bytes, **+0.1%**. Alongside ratings the whole eager path goes 251,518 to 251,604 B gzipped, **+0.03%**.

Switching terms costs one fetch of 17 to 67 KiB gzipped, once per term per session, on an interaction rather than on the eager path. A term already loaded is free.

For the record, the combined single-file version measured earlier in this PR was 157,408 B gzipped eagerly, so the split takes **54.7%** off that.

## Loading, and what `seatsFor` does while a term is in flight

It returns null, the same as any other unknown. The alternative is answering out of whichever term happens to be in memory, which is the confidently wrong number #23 was written to prevent. Seats are only ever read under the term key asked for, so there is no path by which Spring can be served Autumn's numbers.

Null on its own would collapse two different states, so there is a synchronous status:

```
seatsStatus(term)  "ready"    seats are in memory and seatsFor will answer
                   "loading"  a fetch is in flight, ask again when loadSeats resolves
                   "missing"  the index has no such term, Barrett publishes nothing for it
                   "unknown"  the index is not loaded yet, or the term was never asked for
```

"missing" is settled and "unknown" is not, which is the difference between telling someone there are no seat counts for Summer and saying nothing yet.

## Signatures

```js
loadSeats(term, url = "data/seats.json")  // async, resolves to the index
seatsFor(classNumber, term)               // sync, unchanged, null when unknown
seatsStatus(term)                         // sync, "ready" | "loading" | "missing" | "unknown"
seatsTerm(term)                           // sync, the term code once ready, else null
seatsTerms()                              // sync, every term the index lists, oldest first
seatsUpdated(term)                        // sync, that term's Barrett date, from the index
```

`loadSeats()` with no term loads only the index, 361 bytes gzipped, so an early warm call stays cheap. It is safe to call repeatedly, fetches each file once, and a failed fetch can be retried by calling again. `seatsUpdated(term)` works as soon as the index is in, before that term's seats arrive.

`seatsFor(classNumber, term)` keeps its signature and its synchronous null-returning contract, so `render.js`, `filters.js` and `calendar.js` need no change.

## What is left to wire, in the files I do not own

```
js/app.js:248     loadSeats(term).catch(() => null)   inside the existing Promise.all
js/app.js:342     const dated = seatsTerm(term) && seatsUpdated(term) ? ...
js/app.js:360     loadSeats().catch(...)              unchanged, now warms the index only
js/detail.js:126  if (seatsUpdated(term)) ... formatDate(seatsUpdated(term))
```

`term` is already in scope at each. A term change also needs `loadSeats(term)` before the repaint, or the first paint after a switch shows no seats and the second shows them.

## Kept

- Zero limit still returns null, not an empty section. 1408 Spring, 484 Summer and 1425 Autumn rows have no published limit
- A missing class number is still unknown, never zero, and a term absent from the snapshot returns null rather than another term's numbers
- Byte-identical output for unchanged input. Two consecutive full runs, all four files identical:

```
a97719c469f690c7cdf09ca84137cc0ccb0c18b7a5500bc174d6432d11d64656  data/seats.json
56649fd07410afd1aaa80fa236614c9e6dd3f2938fea249c652b7e53b79783ac  data/seats-1262.json
58164189298e6c94e32a01963453006e929921f9344a9fa9e49794e6e4366f22  data/seats-1264.json
fa376a5d6bf02958e9c44b2d437ba85e3532f3a98f3d58a817b8978ec3fd7853  data/seats-1268.json
```

- The residue check still fails a line rather than shipping it. Verified by feeding `parseSubjectFile` a line with one stray character in an unmapped column: 1 parsed, 1 failure, which is what feeds the 0.5% job-failing rate. Nothing is written until every term has cleared its checks, so a bad run cannot leave one term fresh and another stale

43 assertions on `js/seats.js` against the real snapshot, each scenario importing a fresh module so one test's cache cannot answer another's: cold, index-only, mid-flight, loaded, cross-term, retry after a failed index, and the row shapes.

## The commit step, with several outputs

```sh
git add -A -- data/seats.json 'data/seats-*.json'
git diff --cached --quiet && exit 0
```

Staging a file whose bytes did not move records nothing, so a term that did not change stays out of the commit, and `-A` picks up both a new term's file and a dropped one. Verified three ways on the real files:

- a full rerun over an already committed snapshot stages nothing at all
- editing one Spring number and staging lists `M data/seats-1262.json` and nothing else
- `node scripts/fetch-seats.mjs 1264` refreshes only Summer, carries Spring and Autumn through the index untouched, and still stages nothing when the numbers have not moved

A term that stops being searchable is dropped from the index and its file removed, which the same pathspec stages as a deletion. Confirmed with a planted `data/seats-1260.json`.

## The API also reports one thing worth knowing

Barrett rebuilds a live term nightly but freezes a term once it is over. On 2026-08-18 Autumn read `updated: 18-Aug-2026`, Summer `29-Jul-2026` and Spring `27-Apr-2026`. So there is no single date for the snapshot, and that is why `seatsUpdated()` takes a term. Stamping Autumn's date onto Spring's numbers would claim a freshness they do not have.

## Also in here

`seats.yml` picks up the `git pull --rebase` and `push origin HEAD:` that `ratings.yml` and `courses.yml` already do. It was the only one of the three that pushed straight, which fails if anything lands on the branch mid-run, and three jobs now commit to the same branch on their own schedules. The concurrency group is unchanged.


EnesYilmazcode (commented):
**Review: changes requested. The work is right, the packaging is not, and you already said so.**

The parsing, the determinism, the guarantees and the workflow fix all hold. Verified the sizes independently:

```
ratings.json       176 KiB gz
seats.json  now     69 KiB gz
seats.json  PR     153 KiB gz     cold load 245 -> 329 KiB, +34%

  1262   68 KiB gz   updated 2026-04-27
  1264   17 KiB gz   updated 2026-07-29
  1268   70 KiB gz   updated 2026-08-18
```

## The freshness finding justifies the whole shape

Barrett freezing a term's files once the term ends, so the dates genuinely differ per term, is the detail that matters most here. A single file-wide date would have stamped Autumn's 2026-08-18 onto Spring's numbers, which is the same class of defect as #23's term guard and #31's zero limits: presenting a value that looks right in a place it does not apply. Requiring a term on `seatsUpdated()` is the correct response.

## But the packaging should be per-term files

You flagged +34% on every cold load as uncomfortable, and you were right to. 84 KiB to carry two terms most visitors will never select is a real regression, and it lands on the eager path next to ratings, so it is paid before anything renders.

The structure in this PR already nests cleanly by term at 68, 17 and 70 KiB, so the split is a repackaging rather than a redesign. Doing it now, while this is fresh, is cheaper than doing it later once three call sites depend on the combined shape.

Please change it to one file per term, `data/seats-1268.json` and so on, loaded on demand. Cold load returns to today's 245 KiB and a term switch costs 17 to 70 KiB.

Details to get right:
- `loadSeats(term)` becomes term-aware. `js/app.js` already awaits `loadSeats()` inside a `Promise.all` before rendering, so a term-scoped load fits the existing flow. I will do the `app.js` and `detail.js` wiring, you own `js/seats.js` and below.
- `seatsFor(classNumber, term)` must keep its signature and its synchronous null-returning contract, since `render.js`, `filters.js` and `calendar.js` all call it during render and none of them can await.
- Switching term must not leave the previous term's numbers on screen while the new file loads. Decide what `seatsFor` returns in that window and say which you chose. Returning null is fine and honest; returning stale numbers is not.
- Keep a small index of which terms exist so the UI can tell "not loaded yet" from "no data for this term".
- Determinism and the residue check stay as they are.

Also confirm the workflow still commits only changed files when there are several outputs rather than one.

Everything else in this PR stands. Report the new cold-load figure and the per-switch cost.

EnesYilmazcode (commented):
**Approving.** The repackaging landed exactly where it needed to.

```
main today (1 term)      71,265 B gz
PR (index + Autumn)      71,351 B gz
delta                       +86 B, +0.1%
```

86 bytes to gain two more terms, against the +84 KiB the combined version cost. The index at 361 B gzipped is what makes it work: it carries enough to answer `seatsUpdated(term)` and to distinguish a term that has no data from one not yet fetched, without carrying any seats.

Term switching moves from the eager path to an interaction, 17 to 67 KiB once per term per session. That is the right place for it, since most visitors never switch.

## The state model is the part worth keeping

`seatsStatus(term)` returning `ready | loading | missing | unknown` rather than folding everything into null is the correct call. "Barrett publishes nothing for this term" and "this term has not downloaded yet" look identical to `seatsFor` and are completely different facts, and collapsing them is how a loading state ends up rendered as an empty one.

Keeping `seatsFor` synchronous and null-returning was non-negotiable because `render.js`, `filters.js` and `calendar.js` all call it mid-render with no way to await. Verified the signature is unchanged.

Returning null during a load rather than the previously loaded term's numbers is right, and the reasoning is the important bit: seats live in a Map keyed by term and are only read under the key asked for, so there is no code path where Spring can be served Autumn's data. That is a structural guarantee rather than a check that could be forgotten, which is a better answer to #23 than #23 itself had.

## Verified
- Per-term freshness carried into the index, so Spring reads 2026-04-27 rather than inheriting Autumn's 2026-08-18. Barrett freezing a term once it ends is a real behaviour and this is the right response to it.
- Determinism across four files, byte-identical on repeat runs.
- The workflow stages `data/seats.json` and `data/seats-*.json` and exits when nothing moved, including the case where a term stops being searchable and its file is deleted.
- The earlier `pull --rebase` fix is still in.

Merging, then wiring `app.js` and `detail.js` to the new signatures.

PR #49, opened 2026-08-19
Add a test suite

### LINKED ISSUE #37: Add a test suite
There are zero tests. Every bug in v1 and v2 was caught by a hand-written throwaway probe that was then deleted, so nothing protects against reintroducing any of them.

The pure logic is worth locking down, because it is where the real bugs were:

- `js/rank.js` — `parseQuery`, `mergeCourses`, `rankCourses`, `filterCourses`, `isIndividualStudy`
- `js/format.js` — `formatDays`, `formatWhen`, `formatUnits`, and the instructor dedupe in `instructorsOf`
- `js/filters.js` — `toMinutes`, `applyFilters`, and the rule that unknown data never fails a filter
- `js/ratings.js` — `nameKey`, `ratingFor`, including the surname fallback and its refusal to guess
- `js/seats.js` — `seatsFor`, especially the term guard and the zero-limit rule
- `js/calendar.js` — `buildSlots`, especially collapsing sections that share a time

## Regression tests for bugs that actually shipped

Each of these was a real defect caught in review, so each deserves a test that fails if the fix is reverted:

- Empty query must send `q=` rather than omitting it, since the upstream API returns 500 otherwise (#12)
- Grouping by instructor must preserve the section count exactly (#15)
- `surname()` must skip generational suffixes, so `Ivan C. Smith III` sorts under Smith (#15)
- `filterCourses` must never lose a course: primary plus related always equals the full ranked set (#17)
- `seatsFor` must return null for a zero limit rather than reporting `0/0` as open (#31)
- `seatsFor` must return null when the snapshot term does not match the requested term (#23)
- `buildSlots` must collapse sections sharing a day pattern and time into one slot (#33)

## Constraints

- Node's built in `node:test` and `node:assert`. No npm dependencies, matching the rest of the project.
- Pure functions only. Anything touching the network or the DOM is out of scope.
- Fixtures rather than live API calls, so the suite is deterministic and runs offline. Small hand-written fixtures are fine, and preferable to a large captured blob.
- A workflow running the suite on push and on pull request.

## Acceptance

- `node --test` passes from a clean clone
- Every regression listed above has a named test
- Deliberately reverting any one of those fixes makes its test fail. Verify this rather than assuming it, and report which ones you actually checked.

EnesYilmazcode:
Closes #37

133 tests, 315 assertions, no dependencies, `node:test` and `node:assert` only. Runs offline against small hand-written fixtures.

```
$ node --test          (from a clean clone of this branch)
# tests 133
# pass 133
# fail 0
# duration_ms 307
```

| file | tests | asserts | covers |
| --- | --- | --- | --- |
| `tests/rank.test.js` | 25 | 59 | `parseQuery`, `mergeCourses`, `rankCourses`, `filterCourses`, `isIndividualStudy` |
| `tests/filters.test.js` | 20 | 47 | `toMinutes`, `isActive`, `applyFilters`, the unknown-data rule |
| `tests/api.test.js` | 17 | 46 | the request built, the errors returned, `termCodeFor`, `defaultTerm`, paging |
| `tests/format.test.js` | 16 | 39 | `formatDays`, `formatTime`, `formatWhen`, `formatPlace`, `formatUnits`, `instructorsOf` |
| `tests/ratings.test.js` | 16 | 36 | `nameKey`, `surnameKey`, `ratingFor` and its refusals |
| `tests/seats.test.js` | 13 | 26 | `seatsFor`, the term guard, the zero-limit rule |
| `tests/calendar.test.js` | 10 | 24 | `buildSlots` |
| `tests/render.test.js` | 9 | 17 | `groupByInstructor` only, the rest of the file needs a document |
| `tests/courses.test.js` | 7 | 21 | `subjectsFor`, `subjectLabel`, `coursesFor`, `codeFromInput` |

## The reverts, which is the part that matters

A regression test that passes against the broken code is worthless, so I put the bug back for each of the seven and watched the test fail. All seven were checked, not four.

**#12, empty query must send `q=`.** Changed `q: q ?? ""` back to `q: q || undefined` in `js/api.js`.
```
not ok 1 - regression #12: an empty query still sends q=
  error: 'q is missing for ""'
```

**#15, grouping must preserve the section count.** Filed each section under every instructor instead of under the whole set.
```
not ok 6 - regression #15: grouping preserves the section count exactly
  error: |-
    no section added or lost

    8 !== 6
```
Six sections became eight. Four other tests fell over with it, which is what a co-taught section counted twice looks like from every angle.

**#15, `surname()` must skip generational suffixes.** Replaced the backwards scan with `parts[parts.length - 1]`.
```
not ok 7 - regression #15: sorting skips generational suffixes
  + actual - expected
    [
  +   'Ivan C. Smith III',
      'Bob Roberts',
  -   'Ivan C. Smith III',
      'Ann Taylor'
    ]
```
Sorted under I for III, exactly the reported bug.

**#17, `filterCourses` must never lose a course.** Dropped `...setAside` from the related list on the exact-match path.
```
not ok 24 - regression #17: filterCourses never loses a course
  error: |-
    CSE 2321: primary 1 + related 3 != ranked 6

    4 !== 6
```
Two independent study courses vanished, which is the failure #17 described catching by hand.

**#31, `seatsFor` must return null for a zero limit.** Removed `if (limit <= 0) return null;`.
```
not ok 8 - regression #31: a zero limit is unknown capacity, not an empty section
  + actual - expected
  + { enrolled: 0, full: false, limit: 0, waitlist: 1 }
  - null
```
`full: false` on a section with someone already on the waitlist, rendered as `0/0`.

**#23, `seatsFor` must return null on a term mismatch.** Cut the guard down to `if (!snapshot) return null;`.
```
not ok 9 - regression #23: seatsFor returns null when the snapshot is for another term
  error: |-
    Spring 2026 must not read Autumn seats
    + actual - expected

    + { enrolled: 30, full: false, limit: 40, waitlist: 0 }
    - null

not ok 10 - regression #23: seatsFor returns null when no term was asked for
```
Autumn numbers reported against a Spring section, which is the confident lie the guard exists to prevent.

**#33, `buildSlots` must collapse a shared time.** Added `classNumber` to the slot id.
```
not ok 1 - regression #33: sections sharing a day pattern and time collapse into one slot
  error: |-
    Expected values to be strictly equal:

    3 !== 1
```
Three slots where the MoWeFr 3:00p block should be one, the sliced column the feature was built to avoid.

`js/` was restored with `git checkout` after each one and the full suite was re-run green afterwards. Nothing under `js/` is touched by this branch.

## How the loaded modules are tested

`ratings.js`, `seats.js` and `courses.js` keep their data in module-level state that only their loader can fill, and the loaders read `fetch`. Rather than hand-build the index and skip the code that builds it, `tests/helpers.js` serves a small fixture over a stubbed `fetch` and lets the real loader run, so `loadRatings` really does build the `byKey` and `bySurname` maps that `ratingFor` then queries. A test that needs a different snapshot, or none at all, imports a fresh instance with a query string on the specifier.

`api.js` is tested the same way. The stub records the URL, so the `q=` regression is an assertion about the request the client builds, not about a live response. Nothing in the suite opens a socket.

## What is deliberately not covered

Everything that needs a `document`: `renderCalendar`, `renderSlot`, `renderCourse`, `renderSection`, `renderResults`, all of `detail.js` and `app.js`. `groupByInstructor` is exported and pure, so it is tested; the rest of `render.js` is not.

Worth saying plainly: the measured pixel constants in `calendar.js` are the one thing a unit test can never protect. The clipping bug in #33 was invisible to a DOM-presence check and only showed up in a rendered page, and this suite would not catch it coming back. Re-measuring after a type scale change is still a manual step.

## Two things noticed while reading, neither fixed here

Both are behaviour as written, and both now have a test that locks in what the code actually does rather than what the comment implies.

- A minimum rating filter is the one place where unknown data does fail a filter. An unrated instructor scores `-1` against any minimum, so `rating: "1"` hides every section nobody has rated, which duplicates `ratedOnly`. Defensible, but it is not what the "unknown data never fails a filter" comment above `keepSection` says.
- `formatUnits({ minUnits: null, maxUnits: 3 })` renders `null–3 credits`. I did not find a course with that shape in the snapshot, so it may be unreachable, and it is untested rather than fixed.

No real bugs found otherwise. Every regression behaved exactly as its module documents.

## Workflow

`.github/workflows/test.yml` runs `node --test` on Node 22 for every push and pull request. No install step, since there is nothing to install. YAML parses.

`package.json` is new and holds only `"type": "module"` plus a `test` script. Without it Node reads `js/*.js` as CommonJS and every `export` is a syntax error. It adds no dependencies and the browser never sees it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,23 @@
+name: Tests
+
+on:
+  push:
+  pull_request:
+
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+
+      # node:test and node:assert only, so there is nothing to install.
+      - name: Run the suite
+        run: node --test
--- a/package.json
+++ b/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "finder",
+  "version": "0.0.0",
+  "private": true,
+  "description": "OSU course search that leads with the instructor",
+  "type": "module",
+  "scripts": {
+    "test": "node --test"
+  }
+}
--- a/tests/api.test.js
+++ b/tests/api.test.js
@@ -1,0 +1,198 @@
+// The client is only tested for the request it builds and the errors it turns
+// responses into. fetch is replaced, so nothing here touches the network.
+
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { ApiError, fetchTerms, searchClasses, searchAllPages, termCodeFor, defaultTerm } from "../js/api.js";
+
+/** Record every request and answer each with the given body. */
+function capture(body, init = {}) {
+  const calls = [];
+  globalThis.fetch = async (url) => {
+    calls.push(new URL(String(url)));
+    return {
+      ok: init.ok ?? true,
+      status: init.status ?? 200,
+      json: async () => {
+        if (init.badJson) throw new SyntaxError("Unexpected token");
+        return typeof body === "function" ? body(calls.length) : body;
+      },
+    };
+  };
+  return calls;
+}
+
+const original = globalThis.fetch;
+test.afterEach(() => { globalThis.fetch = original; });
+
+const searchBody = { data: { totalItems: 2, totalPages: 1, courses: [{ course: { subject: "CSE" }, sections: [] }] } };
+
+// Regression, #12. Upstream dereferences q without checking it exists, so
+// leaving it off returns a 500 rather than an unfiltered search.
+test("regression #12: an empty query still sends q=", () => {
+  const cases = ["", null, undefined];
+  return Promise.all(
+    cases.map(async (q) => {
+      const calls = capture(searchBody);
+      await searchClasses({ q, term: "1268" });
+      const url = calls[0];
+      assert.ok(url.searchParams.has("q"), `q is missing for ${JSON.stringify(q)}`);
+      assert.equal(url.searchParams.get("q"), "");
+      assert.ok(url.search.includes("q="), url.search);
+    })
+  );
+});
+
+test("searchClasses sends the campus, term and page alongside the query", async () => {
+  const calls = capture(searchBody);
+  await searchClasses({ q: "CSE 2221", term: "1268", page: 3 });
+  const url = calls[0];
+  assert.equal(url.origin + url.pathname, "https://content.osu.edu/v2/classes/search");
+  assert.equal(url.searchParams.get("q"), "CSE 2221");
+  assert.equal(url.searchParams.get("campus"), "col");
+  assert.equal(url.searchParams.get("term"), "1268");
+  assert.equal(url.searchParams.get("p"), "3");
+});
+
+test("searchClasses defaults to page one and reports the totals", async () => {
+  capture(searchBody);
+  const result = await searchClasses({ q: "CSE", term: "1268" });
+  assert.equal(result.page, 1);
+  assert.equal(result.totalItems, 2);
+  assert.equal(result.courses.length, 1);
+});
+
+test("searchClasses refuses to search without a term", async () => {
+  capture(searchBody);
+  await assert.rejects(() => searchClasses({ q: "CSE" }), ApiError);
+});
+
+test("an error status becomes a readable ApiError carrying the status", async () => {
+  capture(searchBody, { ok: false, status: 503 });
+  await assert.rejects(
+    () => searchClasses({ q: "", term: "1268" }),
+    (err) => {
+      assert.ok(err instanceof ApiError);
+      assert.equal(err.status, 503);
+      assert.match(err.message, /503/);
+      return true;
+    }
+  );
+});
+
+test("a body that is not JSON becomes an ApiError", async () => {
+  capture(searchBody, { badJson: true });
+  await assert.rejects(() => searchClasses({ q: "", term: "1268" }), ApiError);
+});
+
+test("a body with no data envelope becomes an ApiError", async () => {
+  capture({ error: "nope" });
+  await assert.rejects(() => searchClasses({ q: "", term: "1268" }), ApiError);
+});
+
+test("a failed connection becomes an ApiError rather than a TypeError", async () => {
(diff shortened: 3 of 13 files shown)


EnesYilmazcode (commented):
**Approving.** This is the most valuable PR in the project so far, and the reason is the revert verification rather than the coverage.

Doing all seven rather than the four I asked for, and reporting the actual failure message each produced, is the difference between "there are tests" and "the tests bite". The #33 one is a good example: adding `classNumber` back into the slot id produced `3 !== 1`, which is precisely the three-way slot slicing that shipped and had to be caught by a rendered screenshot.

133 tests and 315 assertions from a clean clone with no dependencies and nothing opening a socket. `package.json` being required only for `"type": "module"` is a fair call and adds nothing.

Letting the real loaders run against a stubbed fetch, rather than hand-building the `byKey` and `bySurname` indexes, is the right way to test the stateful modules. A test that constructs the index itself would pass even if `loadRatings` stopped building it correctly.

## Finding 1 is a real bug and I have confirmed it

You flagged that the minimum-rating filter is the one place unknown data does fail a filter, contradicting the comment directly above it. It reproduces:

```
baseline        Ali Alilooee | Nickalaus Painter 4.8 | Nabhan Salih 2.7
rating >= 3     Nickalaus Painter 4.8        sections 8 -> 2
rating >= 4     Nickalaus Painter 4.8
```

Alilooee is unrated and disappears at a 3.0 threshold, alongside Salih who is legitimately below it. The `reduce` seeds at `-1`, so an instructor with no rating scores below every threshold.

This matters more than it looks. Only 36% of instructors have any rating, so setting a minimum quietly removes most of the catalogue, and the hidden-count note gives no hint that "unrated" rather than "poorly rated" was the reason. It also makes the rating control silently imply the rated-only checkbox, so two controls that read as independent are not.

Filing and fixing separately, since it needs its own issue and the test you wrote locks in current behaviour and will need updating with it. Correct to leave it alone and report it rather than fixing it inside a test PR.

Finding 2, `formatUnits({minUnits: null})` rendering `null–3 credits`, is worth leaving untested and unfixed given you found no course with that shape. Recorded rather than guessed at.

## Correctly scoped out
Excluding everything needing a `document` rather than faking a DOM, and saying so plainly, is right. So is noting that the measured pixel constants in `calendar.js` are the one thing a unit test can never protect. That bug was invisible to a DOM-presence check and only appeared in a rendered page, and pretending a unit test covers it would be worse than admitting it does not.

Merging.

PR #51, opened 2026-08-19
Stop the minimum rating from hiding unrated instructors

### LINKED ISSUE #50: Minimum rating silently hides unrated instructors
Found by the test suite in #49, confirmed against the live app.

`keepSection()` in `js/filters.js` scores an instructor with no rating as `-1`, because the `reduce` seeds there and an unrated instructor contributes nothing:

```js
const best = people.map((p) => ratingFor(p.name)).filter(Boolean)
  .reduce((max, r) => Math.max(max, Number(r.avgRating) || 0), -1);
if (best < Number(filters.rating)) return false;
```

So every unrated instructor falls below every threshold. Reproduced on CSE 2331:

```
baseline        Ali Alilooee | Nickalaus Painter 4.8 | Nabhan Salih 2.7    8 sections
rating >= 3     Nickalaus Painter 4.8                                      2 sections
```

Alilooee is unrated and vanishes alongside Salih, who is genuinely below the threshold. Those are different reasons and the interface presents them identically.

## Why this matters more than it looks

- **Only 36% of instructors have a rating.** Setting any minimum quietly removes most of the catalogue.
- The hidden-count note says how many sections went, not that "unrated" rather than "poorly rated" was the reason.
- It makes the rating control silently imply "Only rated instructors", so two controls that read as independent are not.
- It contradicts the rule documented in the same file, three lines above the offending code: unknown data never fails a filter. That rule exists so online and arranged sections do not vanish when someone touches a time control, and it should hold here too.

## Fix

A minimum rating should exclude instructors whose **known** rating is below it. An unrated instructor is unknown, not bad, and should survive unless "Only rated instructors" is also ticked. The two controls then compose properly: rating filters quality, rated-only filters presence.

## Note
`tests/filters.test.js` currently locks in the present behaviour, so it needs updating alongside the fix. Do not simply delete the assertion, invert it to assert the corrected rule.

EnesYilmazcode:
Closes #50

Found by #49's test suite within an hour of it landing, which is a reasonable argument for the test suite.

## Before and after, CSE 2331 live

```
                      before                                    after
baseline              Alilooee | Painter 4.8 | Salih 2.7        same
rating >= 3           Painter 4.8                               Alilooee | Painter 4.8
rating >= 4           Painter 4.8                               Alilooee | Painter 4.8
rated only            -                                         Painter 4.8 | Salih 2.7
rated only + >= 4     -                                         Painter 4.8
```

Ali Alilooee is unrated. He used to vanish at a 3.0 threshold alongside Salih, who is genuinely below it. Those are different reasons and the interface presented them identically.

The two controls now compose the way they look like they should: **rating filters quality, rated-only filters presence.**

## Why it happened

```js
.reduce((max, r) => Math.max(max, Number(r.avgRating) || 0), -1)
```

The seed. An instructor with no rating contributes nothing, so the reduce returns `-1`, which is below every threshold. It restores the rule documented three lines above it, which exists so online and arranged sections do not vanish when someone touches a time control.

## Tests
The test that locked in the old behaviour is **inverted, not deleted**, and two were added:
- `ratedOnly` still excludes the unrated, so the controls stay independent
- a known rating below the threshold still fails, so the fix does not just turn the control off

135 tests pass, up from 133.

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -54,9 +54,16 @@ function keepSection(section, filters) {
   if (filters.ratedOnly && !people.some((p) => ratingFor(p.name))) return false;
 
   if (filters.rating) {
-    const best = people.map((p) => ratingFor(p.name)).filter(Boolean)
-      .reduce((max, r) => Math.max(max, Number(r.avgRating) || 0), -1);
-    if (best < Number(filters.rating)) return false;
+    // Unrated is unknown, not bad. Seeding this at -1 made every unrated
+    // instructor fail every threshold, which silently removed most of the
+    // catalogue since only about a third are rated, and made this control
+    // imply the rated-only checkbox. Judge only instructors we have a rating
+    // for; use ratedOnly to exclude the rest.
+    const known = people.map((p) => ratingFor(p.name)).filter(Boolean);
+    if (known.length) {
+      const best = known.reduce((max, r) => Math.max(max, Number(r.avgRating) || 0), 0);
+      if (best < Number(filters.rating)) return false;
+    }
   }
 
   // A section with no meeting pattern has no days to judge, so neither rule
--- a/tests/filters.test.js
+++ b/tests/filters.test.js
@@ -140,11 +140,34 @@ test("a minimum rating compares against the best rated instructor on the section
   assert.equal(applyFilters([cotaught], filters({ rating: "4.5" })).entries.length, 0);
 });
 
-test("a minimum rating does drop a section nobody has rated", () => {
-  // Worth stating plainly, because it is the one place where unknown data does
-  // fail a filter. An unrated instructor scores -1 against any minimum.
+test("regression #50: a minimum rating keeps sections nobody has rated", () => {
+  // Unrated is unknown, not bad. This used to seed the comparison at -1, so
+  // every unrated instructor failed every threshold. With only about a third
+  // of instructors rated, that silently removed most of the catalogue and made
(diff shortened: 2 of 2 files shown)


EnesYilmazcode (commented):
**Approving.** Verified the fix does not simply disable the control, which was the risk.

```
rating >= 3        Alilooee | Painter 4.8         unrated survives, 2.7 drops
rating >= 4        Alilooee | Painter 4.8
rated only         Painter 4.8 | Salih 2.7        unrated drops, both rated stay
rated only + >= 4  Painter 4.8                    composes correctly
```

The middle two rows are the ones that matter. If the fix had gone too far and made the rating check a no-op, `rating >= 4` would still show Salih at 2.7. It does not, so the threshold is still doing its job on data it actually has.

Inverting the existing test rather than deleting it is right, and the two additions cover the failure modes of the fix itself rather than only the original bug. Deleting the old assertion would have left nothing to catch a regression back to `-1`.

## Worth noting about how this was found
The test suite landed and surfaced this within the hour, and it was not found by a test failing. It was found by the agent reading the code closely enough to notice a comment contradicting the three lines beneath it, then writing a test that documented the contradiction rather than papering over it.

That is the second time in this project a comment describing intent has caught a bug the code was silently getting wrong. The rule "unknown data never fails a filter" was written down in #32, and it is what made this visible as a defect rather than a design choice.

Merging.

PR #52, opened 2026-08-19
Reflect a shared query back into the pickers

### LINKED ISSUE #44: Populate the pickers from a shared URL
Opening `?q=CSE%202221&term=1268` runs the search correctly but leaves the subject and number pickers showing their placeholders. A shared link therefore arrives with a rail that looks untouched, even though a course is clearly loaded.

The pickers are a way to build a query rather than a mirror of it, which is why this is not a bug exactly. But someone landing on a shared link sees controls that appear to have nothing to do with what is on screen.

**Change**
- When a query on load parses cleanly as a subject and number, reflect it in the pickers
- Do not force the course index to load eagerly to do it. #36 made that lazy deliberately, so populate on load only if the index is already there, otherwise when it arrives
- A free-text or instructor query should leave the pickers empty, since it does not correspond to a subject and number

**Acceptance**
- `?q=CSE 2221` shows CSE and 2331 in the pickers once the index is available
- `?q=Bucci` leaves them empty
- `courses.json` is still not requested on load

EnesYilmazcode:
Closes #44

```
?q=CSE 2221     subject="CSE"  number="2221"  numberDisabled=false  courses.json=false
?q=Bucci        subject=""     number=""      numberDisabled=true   courses.json=false
?q=CSE (bare)   subject=""     number=""      numberDisabled=true   courses.json=false
no query        subject=""     number=""      numberDisabled=true   courses.json=false
```

The last column is the point. The obvious implementation looks up the query in the course index to confirm the course exists, which would force the 177 KB index onto every shared link and undo #36.

It is not needed. A subject and number are parseable from the query string itself, so the fields fill from `parseQuery` alone and the dropdown options still arrive on first focus. `courses.json` stays absent from the load in all four cases.

Only a query naming **both** a subject and a number maps onto the pickers. `Bucci` is an instructor search and bare `CSE` is a subject-wide search; neither is a course selection, so both correctly leave the rail empty rather than half-filling it.

Also applied on submit and when clicking a name on the landing screen, so the rail tracks the query however it was set rather than only on load.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -1,5 +1,5 @@
 import { fetchTerms, defaultTerm, searchAllPages, ApiError } from "./api.js";
-import { filterCourses } from "./rank.js";
+import { filterCourses, parseQuery } from "./rank.js";
 import { renderResults } from "./render.js";
 import { loadRatings, topRated, ratedCount, profileUrl } from "./ratings.js";
 import { loadSeats, seatsTerm, seatsUpdated, seatsSectionCount } from "./seats.js";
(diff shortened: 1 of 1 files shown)


EnesYilmazcode (commented):
**Approving.**

The interesting decision here is what was *not* done. Confirming the course exists by looking it up in the index is the obvious implementation and would have quietly forced 177 KB onto every shared link, undoing #36 in a change that has nothing to do with loading. Parsing the query instead gets the same result for free.

Verified `courses.json` is absent from the request list in all four cases, including the one where the pickers are populated.

Leaving the pickers empty for `Bucci` and bare `CSE` is right. Half-filling them for a subject-only query would suggest a course had been selected when none had.

Merging.

PR #53, opened 2026-08-19
Make search deterministic with sort=catalogNumber

### LINKED ISSUE #35: Use sort=catalogNumber to make search deterministic
#34 found that OSU's paged search is only non-deterministic under its default relevance sort, and that passing `sort=catalogNumber` makes it exactly repeatable. Verified independently:

```
three identical pulls of subject=cse, term=1268
  no sort              1062, 1064, 1064   intersection 1062   union 1064
  sort=catalogNumber   1064, 1064, 1064   intersection 1064   union 1064
```

`js/api.js` currently works around the old behaviour by fetching five pages and merging, added in #14 to fix instructor search returning nothing for common names. That fix was right for the wrong reason: the problem was never really page count, it was that the pages moved between requests.

**What to change**
- Pass `sort=catalogNumber` from `searchAllPages`
- Re-measure whether five pages are still needed, or whether fewer now suffice, since each page is a request a student waits on
- Do not simply cut the multi-page fetch: `Smith` genuinely has matches beyond page one, and that was the original bug. Measure before trimming.

**Also worth adopting**
`subject=cse` (lowercase, exact) is a real parameter the docs missed. A query that names a subject could use it instead of free text, which would make `CSE 2331` a precise lookup rather than a relevance guess.

`docs/osu-api.md` was updated by #34 already.

EnesYilmazcode:
Closes #35

`sort=catalogNumber` is now passed from `searchAllPages`, but the measurement turned up two things the issue did not anticipate, so the change is not a one-liner. Everything below is measured against the live API for Autumn 2026 (`term=1268`), three or more identical runs each.

## Sorting does fix the paging, by about 98%

Eight identical `q=Smith` pulls, a 674 section result over four pages:

| Order | Sections per pull | In every pull | Sections that moved |
|---|---|---|---|
| relevance | 635 to 674 | 565 | 109 |
| `catalogNumber` | 673 or 674 | 672 | 2 |

Repeating whole five page walks three times:

| Query | relevance | `catalogNumber` |
|---|---|---|
| `Lee` | 435, 443, 443 sections, differs | 464, 464, 464, identical |
| `organic chemistry` | 969, 979, 969, differs | 1000, 1000, 1000, identical |
| `MATH` | 936, 963, 941, differs | 1000, 1000, 1000, identical |
| `CSE` | 928, 1000, 1000, differs | 1000, 1000, 1000, identical |
| `Chen` | 672 every time | 672 every time |

It is not perfect. Two sections still came and went across eight sorted `Smith` pulls, so the doc says a stable order is an observation and not a promise.

## Sorting is not free when the result gets truncated

This is the part I did not expect. A sorted page 1 is the lowest catalog numbers, not the best matches, and `searchAllPages` reads five pages rather than all of them. Distinct courses the query should find, within five pages:

| Query | Pages | Relevance | `catalogNumber` | All of them |
|---|---|---|---|---|
| `MATH` | 11 | 89 | **21** | 89 |
| `creative writing` | 9 | 3 | 3 | 3 |
| `CSE` | 7 | 102 | 101 | 104 |
| `organic chemistry` | 6 | 13 | 13 | 13 |
| `machine learning` | 4 | 5 | 5 | 5 |
| `Smith` | 4 | 86 to 93 | 92 to 93 | 93 |

MATH 1151 alone has 173 sections, so a thousand sections in catalog order never reach the 2000s. Sorting `MATH` blindly would have dropped 68 of its 89 courses.

So the client sorts, reads `totalPages` off the first response, and keeps the sorted order only when the whole result fits in the budget. When it does not fit it pulls the relevance pages too and merges. `mergeCourses` already dedupes by class number, so the sorted page it has already paid for becomes extra coverage rather than waste.

## Five pages is still the right budget

Re-measured, and the answer is no, do not cut it. Distinct real matches by depth:

| Query | Pages | p1 | p2 | p3 | p4 | p5 | All |
|---|---|---|---|---|---|---|---|
| `Smith`, relevance | 4 | 0 | 0 | 57 | 93 | 93 | 93 |
| `Smith`, sorted | 4 | 4 | 14 | 62 | 92 | 92 | 93 |
| `Chen`, relevance | 4 | 0 | 0 | 63 | 87 | 87 | 87 |
| `Chen`, sorted | 4 | 21 | 66 | 74 | 87 | 87 | 87 |
| `Lee`, sorted | 3 | 79 | 168 | 192 | 192 | 192 | 192 |
| `CSE`, relevance | 7 | 44 | 65 | 83 | 95 | 102 | 104 |
| `organic chemistry` | 6 | 11 | 13 | 13 | 13 | 13 | 13 |
| `Bucci` | 1 | 9 | 9 | 9 | 9 | 9 | 9 |

The #14 bug is exactly as real as it was. `Smith` and `Chen` still have zero of that professor's courses on the first two pages under relevance, and are only complete on page 4. The budget stays at five. It is rarely spent in full: rare surnames are one page and common ones are three or four.

## Adopting `subject=`, but only when the query names a number too

`q` turns out to be a union, not an intersection. `q=CSE 2331` matches anything with `CSE` **or** `2331`, which is why it costs seven pages across nine subjects.

| Request | Items | Pages |
|---|---|---|
| `q=CSE 2331` | 1248 | 7 |
| `subject=cse&q=2331` | 50 | 1 |

Checked over 18 course lookups, the narrow form returns the identical course records and the identical set of class numbers as the old five page pull, section for section, nothing missing. All 102 sections of PHYSICS 1250, all 107 of CHEM 1210, all 74 of MATH 1151. What it drops is the thousand sections nobody asked for.

The catch is that `CSE` and `Smith` are the same shape to a regular expression, and `subject=smith` returns zero rather than an error, so a wrong guess wastes a round trip on the commonest search there is. So a token is only read as a subject when the query **also** carries a catalog number, which a bare surname never does, and it falls back to plain text if the subject is not offered. `Smith 2331` takes that fallback and costs one extra request. Bare `MATH` is deliberately left as free text.

## Latency and requests

Six interleaved repeats per query, medians:

| Query | Before | After | Change |
|---|---|---|---|
| `CSE 2331` | 5 req, 572 ms | 1 req, 101 ms | -82% |
| `MATH 1151` | 5 req, 551 ms | 1 req, 139 ms | -75% |
| `PSYCH 1100` | 4 req, 363 ms | 1 req, 117 ms | -68% |
| `PHYSICS 1250` | 5 req, 453 ms | 2 req, 278 ms | -39% |
| `CSE` | 5 req, 445 ms | 6 req, 369 ms | -17% |
| `Bucci` | 1 req, 67 ms | 1 req, 65 ms | -3% |
| `Smith` | 4 req, 379 ms | 4 req, 405 ms | +7% |
| `organic chemistry` | 5 req, 399 ms | 6 req, 434 ms | +9% |
| `machine learning` | 4 req, 343 ms | 4 req, 398 ms | +16% |
| `MATH` | 5 req, 469 ms | 6 req, 558 ms | +19% |
| **total** | **43 req, 4041 ms** | **32 req, 2864 ms** | **-26% requests, -29% time** |

Queries that run past five pages pay one extra request for the sorted first page before the code learns it has to fall back. Course lookups more than pay it back.

## Regression check

37 queries, three runs each, compared through the real `filterCourses`.

- `Smith` returns **93 courses** taught by a Smith, in 7 of 8 sorted runs and 92 in the eighth. Unsorted it ranged 86 to 93. The `primary` list is 25 either way and stopped flapping.
- `Bucci` 9 courses, `Zweben` 3, `Kline` 11, `Soundarajan` 12, `Chen` 87, `Lee` 192. All unchanged or better. `Lee` was 177 to 192 before and is a flat 192 now.
- `CSE 2331` 1 course, 8 sections, identical to before. `MATH 1151` 1 course, 74 sections, identical.
- Across the 18 course lookups: zero sections missing, zero extra.
- Across every non-subject query: zero courses lost against the union of three runs of the old code.
- `primary` results flapped on 6 of 37 queries before and 2 of 37 after. Both remaining are the deliberate relevance fallback for results over five pages.

Edge cases smoke tested: empty query, whitespace, `cse 2331` lowercase, `CSE 2231.01`, `CSE 2331 Bucci`, `ZZZZ 9999`, `Smith 2331`, bare `2331`, and a missing term still raising `ApiError`.

## Also in here: the stale Barrett numbers

Asked for separately, and it touches the same file. The Barrett comparison table in `docs/osu-api.md` came from a hand rolled fixed column slice at columns 20 to 30, which swallows the campus code on regional campus rows, so a line like `CSE 2111  NWK  4816 L` never parsed. Re-measured with `parseSubjectFile` from `scripts/fetch-seats.mjs`, which is the column map #19 verified:

| | Was | Now |
|---|---|---|
| Barrett sections, CSE | 468 | **484**, zero parse residue, all 484 present in `data/seats.json` |
| API sections, CSE | about 1000, varies per pull | **1064**, identical on three full sorted walks |
| Shared by both | 388 | **407** |
| Only Barrett has | about 80 | **77**, of which 16 are regional campus rows |
| Only the API has | about 580 | **657** |
| Barrett names an instructor where the API is blank | 0 | **0**, unchanged |

Every figure in that table was affected, not just the 468, because they all came off the same parse. The headline conclusion survives: Barrett still does not surface a single instructor the API is missing. Two shared sections do carry something in Barrett's instructor column where the API has nobody, but it is `{7W2}`, a session code rather than a person, and the doc now says so. The section also gained a line saying per-section counts must come from the real parser rather than a hand rolled slice, and why.

Barrett's 337 subject files were re-counted and that number was already right.

## Scope

`js/api.js` and `docs/osu-api.md` only. `js/rank.js` needed no change: `mergeCourses` already dedupes sections by class number, which is what lets the sorted and relevance passes be merged safely.

### DIFF
--- a/docs/osu-api.md
+++ b/docs/osu-api.md
@@ -52,12 +52,12 @@ The main endpoint.
 
 | Parameter | Notes |
 |---|---|
-| `q` | Free text. Matches subject, catalog number, course title, and instructor name |
+| `q` | Free text. Matches subject, catalog number, course title, and instructor name. Multi-word queries are a union, not an intersection: `CSE 2331` matches anything with `CSE` or `2331` |
 | `campus` | `col` for Columbus |
 | `term` | A `strm` code, for example `1268` |
 | `p` | 1-based page number |
 | `sort` | `catalogNumber`, `subject`, or `-` prefixed for descending. Default is relevance, which is the source of the paging trouble below |
-| `subject` | Subject code, lowercase. Exact, unlike putting the code in `q` |
+| `subject` | Subject code, lowercase. Exact, unlike putting the code in `q`. Combines with `q` as an intersection |
 
 Neither `sort` nor `subject` needs guessing. Every response carries `sort` and
 `filters` arrays, and each entry in them spells out the parameter and value that
@@ -136,8 +136,8 @@ the section's own limit. `facilityCapacity` is the room's capacity and is not th
 same number. So Finder shows enrolled counts and `enrollmentStatus`, and does not
 draw a percent-full meter it cannot honestly compute.
 
-**Paged search is not deterministic.** Pulling the identical query three times
-back to back returns a different set of sections each time:
+**Paged search is not deterministic in its default order.** Pulling the identical
+query three times back to back returns a different set of sections each time:
 
 ```
 run 1: 1001 CSE sections    totalPages=7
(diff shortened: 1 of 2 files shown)


EnesYilmazcode (commented):
**Approving.** The issue I wrote was half wrong and this PR says so, which is the most useful thing in it.

## The trap is real and worse than reported

I told you to apply `sort=catalogNumber`. Applying it blindly would have been a regression, because a sorted page 1 is the lowest catalog numbers rather than the best matches, and `searchAllPages` reads a budget rather than everything. Verified independently, distinct courses found within 5 pages:

```
q='MATH'              totalPages=11   relevance=213   sorted= 55
q='CSE'               totalPages= 7   relevance=102   sorted=145
q='organic chemistry' totalPages= 6   relevance= 97   sorted= 98
```

MATH is the case that matters: sorting would have cut it from 213 courses to 55. MATH 1151 alone has 173 sections, so 1000 sections in catalog order never reach the 2000s. Sorting when the result fits the budget and falling back to relevance when it does not is the correct resolution, and `mergeCourses` already deduping by class number is why `rank.js` needed no change.

## Verified on the branch

```
q=Smith      pages 4/4   courses=257  instructor-matches=110    772ms
q=Chen       pages 4/4   courses=206  instructor-matches=115    294ms
q=CSE 2331   pages 1/1   courses= 11  instructor-matches=  0     69ms
q=MATH 1151  pages 1/1   courses=  7                            168ms
q=MATH       pages 6/11  courses=243                            632ms
```

`CSE 2331` from 5 requests and roughly 570ms to **1 request and 69ms**. Smith holds at 110 matches, so the #14 bug has not crept back, and MATH correctly engages the fallback rather than stopping at 5 sorted pages.

Refusing to cut the page budget was right. Smith and Chen still have zero of that professor's courses on the first two pages.

## `subject=` adopted narrowly, correctly

```
q='CSE 2331'          totalItems=1248  pages=7
subject=cse&q='2331'  totalItems=  50  pages=1
subject=smith         -> 0
```

`q` being a union rather than an intersection explains the whole 1248. And only reading a token as a subject when a catalog number is also present is the right restraint: `CSE` and `Smith` are the same shape to a regex, and `subject=smith` returns 0 rather than erroring, so guessing would waste a round trip on the most common search there is.

## Every number in that table was wrong, not just the one I found

I asked you to fix 468 and to check whether other figures shared the cause. They all did, because they all came off the same bad parse:

```
Barrett CSE sections   468 -> 484
API CSE sections       "about 1000, varies" -> 1064 exact
shared sections        388 -> 407
Barrett-only            80 -> 77   (16 regional-campus rows)
API-only               580 -> 657
```

The conclusion survives: still zero instructors Barrett has that the API lacks. And catching that two of the 407 shared rows have Barrett content where the API is blank, but that the content is `{7W2}`, a session code rather than a person, is exactly the check that stops a corrected number becoming a new wrong claim.

The parser warning is the durable fix. My original count and my first attempt at re-counting it both used a hand-rolled slice, and both were wrong.

Merging.

PR #94, opened 2026-08-20
Fix stats page displaying zeros when no data is available

### LINKED ISSUE #73: Stop the stats page showing zeros when it has no data
The stats page hides its dashboard with the `hidden` attribute, but a bare `main { display: grid }` overrides the browser's own `[hidden] { display: none }`, so a failed load prints "Could not load the numbers" above PAGE VIEWS 0 / VISITORS 0 / COUNTRIES 0. Add `main[hidden] { display: none }` next to the existing rule.

"No traffic" and "no data" are opposite claims and the page currently makes the wrong one. The same rule also means every healthy load flashes a zeroed dashboard before the numbers arrive.

Measured on three branches, a rejected fetch, an HTTP 500 and a held happy path: `#main.hidden === true` while `getComputedStyle(#main).display === "grid"`, with a 1032x513 rect. A walk of `document.styleSheets` finds exactly one matching rule. The mechanism is origin, not specificity: author CSS beats the user-agent sheet either way, and `[hidden]` is the more specific selector of the two. `#status` on the same page uses a class selector and its `hidden` works correctly, which is the control that proves the diagnosis.

Error sentence above a dashboard of zeros
Look at the error line sitting above three tiles reading 0.

## Steps to reproduce

1. Open https://enesyilmazcode.github.io/Finder/stats/ with the analytics host blocked.
2. The error sentence appears and the zeroed dashboard stays under it instead of being hidden.

## Where it is

```css
/* css/stats.css:103 */
main { display: grid; gap: var(--gap); }
```

`<main id="main" hidden>` is `stats/index.html:32`.

## Suggested fix

Add `main[hidden] { display: none }` beside the `main` rule.

Sa7yaC:
closes #73

### DIFF
--- a/css/stats.css
+++ b/css/stats.css
@@ -102,6 +102,8 @@ body {
 
 main { display: grid; gap: var(--gap); }
 
+main[hidden]{ display: none; }
+
 .tiles {
   display: grid;
   grid-template-columns: repeat(auto-fit, minmax(9rem, 1fr));


EnesYilmazcode (approved):
Thanks for the contribution, and for keeping it inside the scope the issue described.

I verified it in headless Chrome on both states. On `main`, `#main` has `hidden === true` while `getComputedStyle(#main).display === "grid"` with a 764x66 rect, which is why the zeroed dashboard renders under the error line. With your rule the same element computes to `display: none` at 0x0, and the healthy path with no `hidden` attribute still computes to `grid` at 764x66, so nothing else moved.

Approving.

